PR #76, opened 2015-01-15
always use current auth when loading xform session

snopoke:
http://manage.dimagi.com/default.asp?153046

Fixes a bug that would give an auth failure message when trying to open an old XForm session for which the django session has expired.


czue:
tricky. nice find!

PR #78, opened 2015-02-11
Update form step API call

wpride:
http://manage.dimagi.com/default.asp?134659#831640

### DIFF
--- a/touchforms/backend/xformplayer.py
+++ b/touchforms/backend/xformplayer.py
@@ -209,17 +209,12 @@ class XFormSession:
 
     def _walk(self, parent_ix, siblings):
         def step(ix, descend):
-            next_ix = self.fem.incrementIndex(ix, descend)
+            next_ix = self.fec.getAdjacentIndex(ix, True, descend)
             self.fem.setQuestionIndex(next_ix)  # needed to trigger events in form engine
             return next_ix
 
         def ix_in_scope(form_ix):
-            if form_ix.isEndOfFormIndex():
-                return False
-            elif parent_ix.isBeginningOfFormIndex():
-                return True
-            else:
-                return FormIndex.isSubElement(parent_ix, form_ix)
+            return form_ix.isInForm()
 
         form_ix = step(parent_ix, True)
         while ix_in_scope(form_ix):
--- a/touchforms/backend/xformplayer.py
+++ b/touchforms/backend/xformplayer.py
@@ -221,7 +221,6 @@ class XFormSession:
             else:
                 return FormIndex.isSubElement(parent_ix, form_ix)
 
-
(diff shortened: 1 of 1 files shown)


touchforms/backend/xformplayer.py
```diff
@@ -209,17 +209,12 @@ def walk(self):
...
             return next_ix
 
         def ix_in_scope(form_ix):
-            if form_ix.isEndOfFormIndex():
-                return False
-            elif parent_ix.isBeginningOfFormIndex():
-                return True
-            else:
-                return FormIndex.isSubElement(parent_ix, form_ix)
+            return form_ix.isInForm()
```
ctsims: the semantics of this don't seem quite the same as before? Beginning of form index should still be valid, right?

wpride: [2/10/15, 11:23:51 PM] William Pride: @ your comment, that’s true but those functions seem to be used the same
[2/10/15, 11:25:00 PM] Clayton Sims: hmmm, might be worth seeing if Simon has anything to say, I'm not 100% sure but the beginning index might need to be in scope for certain circumstances?

@snopoke you can see here

https://github.com/dimagi/javarosa/blob/239e578a3b1049aaddf3f9a03612082ebedf92ef/core/src/org/javarosa/form/api/FormEntryController.java#L261

That we've switched to this isInForm() method for basically the same pattern here. The main semantic difference seems to be that the old method returns true when the parent is the beginning of the form - any particular reason to keep this around? (this wasn't necessary for the fix, but I included to update the pattern)

snopoke: I definitely don't think this is correct since this method gets called recursively for groups and repeats so that `ix_in_scope` isn't just checking if the `form_ix` is in the form but if its in the scope of the current `parent_ix`.

wpride:
@snopoke rolled back that portion - out of curiosity, what case would that have broken? would like to test out

snopoke:
@wspride it would cause an infinite recursive loop when trying to add a new repeat to a repeat group. It would also mess up question groupings (all questions after the first group would be included in it.

czue:
no way, this is awesome if it fixes that thing! does this need QA or you're confident it's good to go?

wpride:
Is there a standard set of forms we run through? I'm pretty new to this
ballgame so "confident" is a bit strong... it works in the few forms I've
tested it with

czue:
i guess more confident would be understanding the difference between those two calls and knowing that the new one is doing the right thing.

czue:
@wpride waiting for you to confirm whether you want merge or want to put a QA label on this.

wpride:
@czue merge

czue:
do it live!

PR #79, opened 2015-02-11
Update libs to 2.19

wpride:
fixes for log()

PR #80, opened 2015-02-13
Revert "Update form step API call"

czue:
Reverts dimagi/touchforms#78

This actually depends on https://github.com/dimagi/touchforms/pull/79


czue:
@wpride

czue:
@wpride can you just add this change to the jar update branch and we can do QA together?

wpride:
Sounds good

PR #84, opened 2015-03-09
Update jars

benrudolph:
@wpride


wpride:
@czue do we want QA for hot fixes or no?

czue:
@benrudolph sanity tested and I ran sms tests. think it's worth moving forward given the ipm issues.

PR #90, opened 2015-04-24
Double submit

benrudolph:
Fixes this: http://manage.dimagi.com/default.asp?114082#925589

Also refactors some of the JS structure, adds tests, and travis


touchforms/formplayer/static/formplayer/script/xforgasm.js
```diff
@@ -16,8 +16,53 @@ var ERROR_MESSAGE = "Something unexpected went wrong on that request. " +
     "If you have problems filling in the rest of your form please submit an issue. " +
     "Technical Details: ";
 
+function TaskQueue() {
```
benrudolph: A bit overkill for the task at hand, but I think we can extend this to refactor the whole BLOCKING_REQUESTS structure

benrudolph:
@czue @millerdev

touchforms/formplayer/static/formplayer/script/test_helpers/jasmine-fixture.js
```diff
@@ -0,0 +1,433 @@
+/* jasmine-fixture - 1.2.2
```
czue: is this just a static library file?

touchforms/formplayer/templates/fullform.html
```diff
@@ -3,10 +3,11 @@
   <head>
     <meta http-equiv="content-type" content="text/html; charset=UTF-8" />
 
-    <script language="javascript" type="text/javascript" src="{{STATIC_URL}}formplayer/script/jquery-1.8.3.min.js"></script>
-    <script language="javascript" type="text/javascript" src="{{STATIC_URL}}formplayer/script/jquery-ui.min.js"></script>
+    <script language="javascript" type="text/javascript" src="{{STATIC_URL}}formplayer/script/vendor/jquery-1.8.3.min.js"></script>
+    <script language="javascript" type="text/javascript" src="{{STATIC_URL}}formplayer/script/vendor/jquery-ui.min.js"></script>
     <link rel="stylesheet" href="http://ajax.googleapis.com/ajax/libs/jqueryui/1.8.5/themes/ui-lightness/jquery-ui.css">
     <script language="javascript" type="text/javascript" src="{{STATIC_URL}}formplayer/script/shortcut.js"></script>
+    <script language="javascript" type="text/javascript" src="{{STATIC_URL}}formplayer/script/vendor/underscore.js"></script>
```
czue: not sure if the HQ version doesn't bother with these templates or whether there's a separate change needed there

czue:
this looks totally awesome!

czue:
happy to merge as-is, not sure if there was something specific you wanted to test before that though.

touchforms/formplayer/static/formplayer/script/test_helpers/jasmine-fixture.js
benrudolph: yea, does our deploy automatically compress? i usually like using the dev version in case i need to step through

czue: i believe we gzip if the client supports it but don't automatically compress unless explicitly put in a `{% compress %}` templatetag.

touchforms/formplayer/templates/fullform.html
benrudolph: so i'm a bit conflicted about this, our HQ does load underscore.js, but i'd like to be able to test the js code in touchforms without having to depend on HQ. maybe a better solution would be to check if underscore has been loaded first though it's unclear if we can do that in the template

touchforms/formplayer/static/formplayer/script/test_helpers/jasmine-fixture.js
benrudolph: cool, probably not need here since it's only loaded in tests, but that's definitely something for the `underscore.js` file

benrudolph:
Ok should be good to go now. It now assumes that when loaded in HQ it has access to underscore. When I test I just load it from a directory. later i'll probably want to change to bower so it's easier to lock the version with HQ's version

czue:
Cool. I'll give @millerdev a chance to comment on js tests and what not, but I'm good with this being merged whenever.

benrudolph:
:+1:

touchforms/formplayer/static/formplayer/script/xforgasm.js
```diff
@@ -16,8 +16,52 @@ var ERROR_MESSAGE = "Something unexpected went wrong on that request. " +
...
+ * Executes the queue in a FIFO action. If name is supplied, will execute the first
+ * task for that name.
+ */
+TaskQueue.prototype.execute = function(name) {
+    var task,
+        idx;
+    if (name) {
+        idx = _.indexOf(_.pluck(this.queue, 'name'), name);
+        task = this.queue.slice(idx, idx + 1)[0];
+        this.queue = [].concat(this.queue.slice(0, idx), this.queue.slice(idx + 1));
```
millerdev: I think this would be more efficiently (and more correctly if a task with the given name is not found) done like this:
```js
idx = _.findIndex(this.queue, function (task) {
    return task.name === name;
});
if (idx !== -1) {
    task = this.queue.splice(idx, 1)[0];
}
```

benrudolph: `splice`! i knew there was a better js built in function i just couldn't seem to find it at the time of writing. thanks i'll update

touchforms/formplayer/static/formplayer/script/xforgasm.js
```diff
@@ -119,6 +163,9 @@ function WebFormSession(params) {
     self.serverRequest = function (params, callback, blocking) {
         var that = this;
         var url = that.urls.xform;
+        if (params.action === 'submit-all' && this.NUM_PENDING_REQUESTS) {
+            self.taskQueue.addTask(params.action, self.serverRequest, arguments, self)
```
millerdev: Is `self` different from `this` and `that`? It's a little confusing to be using all three in this context when `this` would seem to work just fine.

benrudolph: definitely confusing and will be refactored at a later time. for now i'll switch `this.NUM_PENDING_REQUESTS` to `self.` for consistency

touchforms/formplayer/static/formplayer/script/test_vendor/underscore.js
```diff
@@ -0,0 +1,1548 @@
+//     Underscore.js 1.8.3
```
millerdev: I'm not sure if this is only for tests, or if it's used in any other context (looks like only tests). [commcare-hq](https://github.com/dimagi/commcare-hq) currently contains at least (from a quick search) Underscore 1.3.1, 1.4.4 and 1.6.0, and Vellum uses 1.5.2. All of them are pretty old and contain backward-incompatible APIs with 1.8.x. This is a pain for Vellum dev, but I haven't tried to tackle it yet as it seems to be a pretty nasty problem to bring all API use up to the latest version. And maybe we want to consider switching to lodash? Anyway, this is mainly just FYI. Not sure if there is anything actionable other than it could be confusing to have different versions in different parts of the codebase, especially given that there are backward-incompatible changes in newer versions.

benrudolph: would love to switch to lo-dash. yea you're right it's only for testing, i should change it to 1.4.4 since that's what's loaded on the page, hence i can't actually use `findIndex`

benrudolph:
:white_check_mark: comments addressed

touchforms/formplayer/static/formplayer/script/xforgasm.js
```diff
@@ -16,8 +16,51 @@ var ERROR_MESSAGE = "Something unexpected went wrong on that request. " +
...
+/*
+ * Executes the queue in a FIFO action. If name is supplied, will execute the first
+ * task for that name.
+ */
+TaskQueue.prototype.execute = function(name) {
+    var task,
+        idx;
+    if (name) {
+        idx = _.indexOf(_.pluck(this.queue, 'name'), name);
+        task = this.queue.splice(idx, 1)[0];
```
millerdev: what happens here if `idx === -1`?

PR #91, opened 2015-04-24
Move templates into HTML files and use grunt to compile

benrudolph:
In addition to making the templates compile to a single file so we can use JST, this fixes: http://manage.dimagi.com/default.asp?47382#926323 (repeat group UX) by putting the `add new` link at the bottom. Also in the future it'd be nice to have `fullform-ui.templates.js` generated on the server. If it's OK with people I also want to use `grunt watch` to uglify and compile all the src files into one `formplayer.min.js` so we can just include that in `cloudcare_home` instead of the 5 files. I also want to slowly move this to coffee script which grunt can also do via watch. @millerdev had a talk with @emord about using requirejs and i think this repo is a bit far away from using it. In the future though I'd like to see it. Would love to have a conversation about pros and cons about the setup. Would like to align with Vellum as much as possible to make working in either pretty similar. @czue


czue:
(don't feel qualified to review this but like the concept!)

millerdev:
> I also want to slowly move this to coffee script

I don't have any experience with coffee script, so maybe I'm just an old fuddy-duddy. Do we really want to introduce yet another language (dialect?) to our codebase? I think we need to demonstrate a pretty clear advantage for it since there are clear disadvantages to introducing more dependencies.

touchforms/Gruntfile.js
```diff
@@ -0,0 +1,24 @@
+module.exports = function(grunt) {
+
+    // Project configuration.
+    grunt.initConfig({
+    pkg: grunt.file.readJSON('package.json'),
```
millerdev: Seems like this should be indented another level.

touchforms/formplayer/static/formplayer/script/fullform-ui.templates.js
```diff
@@ -0,0 +1,48 @@
+this["JST"] = this["JST"] || {};
+
+this["JST"]["formplayer/templates/formplayer/fullform-ui/form.html"] = function(obj) {
+obj || (obj = {});
+var __t, __p = '', __e = _.escape;
+with (obj) {
+__p += '<div>\n    <h1 id="title"></h1>\n    <div id="form"></div>\n    <input id="submit" type="submit" value="Submit" />\n</div>\n';
```
millerdev: This looks like auto-generated code. Should it reside in the repo?

benrudolph: no, but we don't currently have a build process setup on the server for it. so for now it has to be in the repo in order for things to work

benrudolph:
Sure, happy to have that discussion at some point. i'm not going to introduce it tomorrow, but i think coffeescript is an incredible improvement over javascript. it's not an esoteric dialect that we'd be taking a risk using. this project seems self contained enough that it might be a good demonstration of it and it's already dependent on grunt for templating. ideally there will be one grunt command that will make the templates, process the coffeescript and build into one file

touchforms/formplayer/static/formplayer/script/fullform-ui.templates.js
millerdev: Ugh. Vellum when through a phase like that and it was painful. We even used grunt for a while, but that's another topic. The source code got out of sync with the build artifacts and confusion and buggy deploys ensued. What about making a [deploy script like Vellum has](https://github.com/dimagi/commcare-hq/blob/master/scripts/vellum-to-hq) to prepare a new version of touchforms for deploy? The vellum script is run manually to push a new version of vellum to HQ, and this does not happen on nearly every deploy, but it is the one true way to get a new version of Vellum into HQ.

benrudolph: agreed, and that does sound awful. a separate build does sound reasonable to me. it may even be wise to just put it into our fabfile since touchforms is bit more dependent on HQ?

millerdev: Putting it in the fabfile seems like a nice idea. However, as we already discussed offline, it may be non-trivial considering this depends on npm, and getting npm to run as part of the deploy has its own set of issues.

benrudolph:
@millerdev, i had a discussion about this with @dannyroberts. how about this for a way forward: we merge this for now, realizing that it's not the best to keep the compiled code in the repo. i have on my todo list to integrate a build process in the fabfile. it will look something like this:
1. have ansible install npm on hqdjango servers (not part of fabfile)
2. have deploy process run `grunt compile` in touchforms

one we have that, we can remove this from the repo. right now though it'd be nice to continue to work on some of the changes i've made without having to do a whole bunch of deploy work right now. does that sound reasonable?

millerdev:
> it'd be nice to continue ... without having to do a whole bunch of deploy work right now

I think this is risky because it means you have a sub-optimal deploy process for some undetermined period of time. If you never get around to fixing it and other people make contributions but miss this extra step then they'll be confused at best and we'll have buggy deploys at worst. I also think you may be underestimating how difficult it could be to roll npm into our deploy and make it reliable 100% of the time. I'm not on board yet, but will agree to merge this if others think I'm being overly cautious.

dannyroberts:
@millerdev from my discussion with @benrudolph it seemed that it would be pretty hard to push without the corresponding change to the built file because the testing cycle involves building that file (if you don't, you won't be able to see your changes on your localhost:8000). Does that change anything for you?

benrudolph:
also fwiw, npm is on all the machines already, it's under the common_installs: https://github.com/dimagi/commcarehq-ansible/blob/nginx-redirect-com/ansible/roles/common_installs/tasks/main.yml#L22. i'm curious as to why you're concerned that it'll be hard to run an npm command reliably

millerdev:
> the testing cycle involves building that file ... Does that change anything for you?

@benrudolph had already mentioned that in a previous discussion. Vellum had the same issue (to debug in a browser you had to rebuild/minify). We still saw people forget to build. For example, in the case of one more quick/harmless edit that should fix that bug I saw in my final test round. So simple it doesn't merit testing again... Bad practice, I know, but it happens.

> i'm curious as to why you're concerned that it'll be hard to run an npm command reliably

Because it downloads lots of dependencies from sources outside of our control. Hosted versions of things change without warning. Things that were stable and worked reliably [suddenly break](http://www.reddit.com/r/javascript/comments/2whrl9/underscore_18_makes_last_minute_breaking_changes/). When the person doing the deploy is not intimately familiar with the sub-component, they're in for a real treat when that happens.

One way to make it reliable is to host all of the dependencies on our own server (although ironically we've even had deploy problems when files.dimagi.com goes down). The last time I looked that was a non-trivial thing to do. Don't forget you need to design a process for updating those hosted dependencies.

I'm starting to feel silly making this case because most of the time there will be no problem and you'll say "See it works. What was all that fuss about?" If y'all think I'm being overly cautious then go ahead and merge.

dannyroberts:
:x:

benrudolph:
:white_check_mark:

benrudolph:
;)

PR #94, opened 2015-04-27
restore __init__.py

czue:


czue:
@benrudolph fyi

benrudolph:
ah oops, didn't realize that would affect things

PR #99, opened 2015-05-03
Revert removal of preloader support

NoahCarnahan:
Addresses http://manage.dimagi.com/default.asp?166907
Primarily reverts 3c95c9a6cfaa31891bca99c7cba2e1e7ca05895c (also fixes readme).
I've tested that forms with preloads work, but I'm not certain how to test that I didn't break the ability to register custom function handlers via context. Look ok to you @czue ?
FYI review buddy @biyeun

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -24,16 +24,19 @@ Update your settings.py::
     Add XFORMS_PLAYER_URL = "http://127.0.0.1:4444"
 
 Syncdb::
+
     python manage.py syncdb
 
 Set localsettings.py for this backend::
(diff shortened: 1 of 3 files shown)


czue:
Looks fine. custom function handlers are unused for now anyway, so no worries if they broke - i'll make sure to test again when 2.21 is out.

czue:
(might be nice if preloaders logged a deprecation warning or something so we can remove them later, but seems out of scope for now)

czue:
pinging @benrudolph in case he wants to double-check readme changes

PR #105, opened 2015-05-07
Better log formatting

benrudolph:
@esoergel 
minor formatting change, but will help a lot in getting performance of requests so we can start measuring on prod which requests take a long time

### DIFF
--- a/touchforms/backend/xformserver.py
+++ b/touchforms/backend/xformserver.py
@@ -17,7 +17,11 @@ init_classpath()
 import com.xhaus.jyson.JysonCodec as json
 from xcp import TouchFormsException
 
-logging.basicConfig(stream=sys.stderr, level=logging.DEBUG)
+logging.basicConfig(
+    stream=sys.stderr,
+    level=logging.DEBUG,
+    format='%(asctime)s %(name)-12s %(levelname)-8s %(message)s'
+)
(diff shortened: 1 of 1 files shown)


esoergel:
Looks good.  Not entirely sure how to wire it up end-to-end, but wanted to flag that you can specify various formatters in the logging config dict: https://github.com/dimagi/commcare-hq/blob/master/settings.py#L764-L773  The idea is you'd register a specific logger for xformserver or whatever, then tell it to use this formatter.  This works too though.

benrudolph:
Yea, would be nice to have it there. right now though this is very divorced from anything django. i saw that django 1.7 works with jython. then we might be able to share settings

dannyroberts:
Maybe I'm missing context for why we want to change, but FWIW I think letting touchforms have its own localsettings seems totally reasonable if the concept is to have it be its own independent service, which I think is for the most part already the case.

esoergel:
:+1: I was more thinking about keeping formatting config in *a* localsettings (rather than defining in-line) - definitely in favor of touchforms being an entirely separate service.

benrudolph:
ah got it. yea, would be nice for every log file to have the same format somehow

benrudolph:
:+1:

PR #106, opened 2015-05-07
Use exceptions

benrudolph:
It's pretty painful working with Jython 2.5.3. It means I can't use a lot of nice python things like `with` and modules have been a huge thorn in my side which is why there is no `tests` directory. The Jython classpath gets really messed. Jython 2.7 just got released and I think we should try to upgrade. I was able to more easily create modules and I don't have to think about the differences between python 2.5 and 2.7

@esoergel

### DIFF
--- a/touchforms/backend/test_server.py
+++ b/touchforms/backend/test_server.py
@@ -1,0 +1,63 @@
+import unittest
+import sys
+from xcp import InvalidRequestException
+import xformserver
+import xformplayer
+
+
+class DummyServer(xformserver.XFormHTTPGateway):
+    def __init__(self):
+        pass
+
+
+class XFormServerTest(unittest.TestCase):
+
+    def setUp(self):
+        self.server = DummyServer()
+
+    def test_param_no_session_id(self):
+        content = {
+            'action': xformplayer.Actions.ADD_REPEAT
+        }
+        try:
+            xformserver.handle_request(content, self.server)
+        except InvalidRequestException, e:
+            self.assertTrue('session' in e.message)
+        else:
+            self.fail()
+
+    def test_param_no_answer(self):
+        content = {
+            'action': xformplayer.Actions.ANSWER,
+            'session-id': '123'
+        }
+        try:
+            xformserver.handle_request(content, self.server)
+        except InvalidRequestException, e:
+            self.assertTrue('answer' in e.message)
+        else:
+            self.fail()
+
+    def test_param_no_action(self):
+        content = {}
+        try:
+            xformserver.handle_request(content, self.server)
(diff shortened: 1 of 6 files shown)


czue:
nice cleanup!

touchforms/backend/test_server.py
```diff
@@ -0,0 +1,63 @@
+import unittest
+import sys
+from xcp import InvalidRequestException
+import xformserver
+import xformplayer
+
+
+class DummyServer(xformserver.XFormHTTPGateway):
+    def __init__(self):
+        pass
```
czue: this is clever

touchforms/backend/xformserver.py
```diff
@@ -70,22 +70,20 @@ def do_POST(self):
...
             if isinstance(e, java.lang.Exception):
                 e.printStackTrace()  # todo: log the java stacktrace
             elif isinstance(e, urllib2.HTTPError):
                 if e.headers.get("content-type", "") == "text/plain":
                     msg = e.read()
-            elif isinstance(e, TouchFormsException):
-                error_type = type(e).__name__
 
-            logging.exception('error handling request')
+            info = sys.exc_info()
```
esoergel: What's the advantage of this over extracting information from `e`?

touchforms/backend/xformserver.py
```diff
@@ -141,15 +139,25 @@ def do_GET(self):
...
+def ensure_required_params(params, action, content):
+    for param in params:
+        if param not in content:
+            raise InvalidRequestException("'%s' is required for action: %s" % (param, action))
+
+
 def handle_request(content, server):
     if 'action' not in content:
-        return {'error': 'action required'}
+        ensure_required_params(['action'], 'All actions', content)
```
esoergel: woo! validating user input!
You can probably get rid of `if 'action' not in content:` here, right?

esoergel:
Nice reworking!  Looks g2m when you're ready.

snopoke:
nice, great to have some tests going here that we can build on!

benrudolph:
Thanks for feedback will update. There's actually an error in this so please don't deploy. Touchcare api doesn't require session id

benrudolph:
(Will fix once I get to the office)

touchforms/backend/xformserver.py
benrudolph: I was having trouble getting consistent information from the error message. if i raised a java.lang.Exception i'd have different methods than a regular python exception. How do you normally do it?

esoergel: Sounds legit. I have no idea how one normally handles java exceptions in Jython.

PR #109, opened 2015-05-11
Send 400 response when malformed params

benrudolph:
@esoergel @dannyroberts don't send 500 when params are wrong


esoergel:
Looks good to me.  Know if any tests check that status code?

benrudolph:
None do, i haven't built infrastructure to mock the entire server

benrudolph:
no rush though, can wait

PR #114, opened 2015-05-15
Removed curious code written by Drew

kaapstorm:
It seems Drew wrote a function to return the first non-None parameter, and then only used it in place of dict.get. It doesn't seem to be used anywhere else. I removed it, and just used dict.get.

Am I missing something?

@snopoke @czue @millerdev

### DIFF
--- a/touchforms/formplayer/views.py
+++ b/touchforms/formplayer/views.py
@@ -72,22 +72,16 @@ def download(request, xform_id):
         return response
 
 
-def coalesce(*args):
-    for arg in args:
-        if arg is not None:
-            return arg
-    return None
-
 @csrf_exempt
 def enter_form(request, **kwargs):
     xform_id = kwargs.get('xform_id')
     xform = kwargs.get('xform')
     instance_xml = kwargs.get('instance_xml')
-    session_data = coalesce(kwargs.get('session_data'), {})
-    input_mode = coalesce(kwargs.get('input_mode'), 'touch')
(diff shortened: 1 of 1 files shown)


czue:
I think the behavior is subtly different if the dict has `None` arguments in it.

```python
>>> def coalesce(*args):
...     for arg in args:
...         if arg is not None:
...             return arg
...     return None
... 
>>> 
>>> coalesce({'foo': None}.get('foo'), 'bar')
'bar'
>>> {'foo': None}.get('foo')
```

I don't know what kind of testing went into this, but might be worth reverting if we're not sure it doesn't hit that use case.

cc @snopoke

snopoke:
Good point

touchforms/formplayer/views.py
```diff
@@ -72,22 +72,16 @@ def download(request, xform_id):
...
 @csrf_exempt
 def enter_form(request, **kwargs):
     xform_id = kwargs.get('xform_id')
     xform = kwargs.get('xform')
     instance_xml = kwargs.get('instance_xml')
-    session_data = coalesce(kwargs.get('session_data'), {})
-    input_mode = coalesce(kwargs.get('input_mode'), 'touch')
-    submit_callback = coalesce(kwargs.get('onsubmit'), default_submit)
-    abort_callback = coalesce(kwargs.get('onabort'), default_abort)
-    force_template = coalesce(kwargs.get('force_template'), None)
```
millerdev: This one is silly. All the rest seem legit for the reason that @czue mentioned.

PR #111, opened 2015-05-12
Form context

benrudolph:
@czue 
elf: @esoergel

PR #118, opened 2015-05-20
Use redis for caching

benrudolph:
@czue 
elf: @snopoke 

Was thinking a bit about the cause of all those requests. Had an idea to remove heartbeat completely since all I think it does is ensure that the python dict that holds the sessions doesn't get too large in memory. We can use redis for this which will automatically purge things. This PR isn't completely clean yet, but wanted to get a little feedback before pushing all the way (as many of my optimization attempts have been foiled)


touchforms/backend/xformplayer.py
```diff
@@ -40,77 +41,52 @@
 class NoSuchSession(Exception):
     pass
 
-class global_state_mgr(object):
+class GlobalStateManager(object):
```
dimagimon: E302 expected 2 blank lines, found 1

touchforms/backend/xformplayer.py
```diff
@@ -40,77 +41,52 @@
...
+            xform_state_raw = self.redis.get(session_id)
+            if xform_state_raw:
+                xform_state = json.loads(xform_state_raw)
+                xform_state.update({'uuid': session_id})
+                return XFormSession(**xform_state)
+
+            # see if session has been persisted
+            xform_session = persistence.restore(session_id, XFormSession, override_state)
+            if xform_session:
+                self.new_session(xform_session) # repopulate in-memory cache
```
dimagimon: E261 at least two spaces before inline comment

touchforms/backend/setup.py
```diff
@@ -15,6 +15,7 @@ def init_classpath():
 
     jardir = os.path.join(CUR_DIR, 'jrlib')
     jars = [k for k in os.listdir(jardir) if k.endswith('.jar')]
+    sys.path.append('/Users/benrudolph/.virtualenvs/commcare-hq/lib/python2.7/site-packages')
```
benrudolph: I'll either have to direct this to the python_env in our code root, or start a new virtual env for touchforms and check it in to the repo. i think the later would be the least amount of configuration, but slightly clunky cause you'd have to check in the virtual env...

touchforms/backend/xformplayer.py
```diff
@@ -40,77 +41,52 @@
...
+            db=settings.REDIS_DB,
+        )
+
+    def new_session(self, xform_session):
         with self.lock:
-            self.session_cache[xfsess.uuid] = xfsess
-            self.ctx.setNumSessions(len(self.session_cache))
+            state = xform_session.session_state()
+            self.redis.set(xform_session.uuid, json.dumps(state))
+            #self.ctx.setNumSessions(len(self.session_cache))
```
benrudolph: Not sure what to do about this guy. Seems like this is only used for the offline mode. I don't think I've ever seen the offline mode

benrudolph:
also this was really hard to make automated tests for since XFormSession is so heavily coupled with making calls to HQ

touchforms/backend/xformplayer.py
czue: (I think we can kill offline support for now)

touchforms/backend/setup.py
snopoke: checking in the virtual env doesn't sound great, also are virtual envs portable like that?

touchforms/backend/test_global_state_manager.py
```diff
@@ -0,0 +1,103 @@
+import os
+import unittest
+import xformplayer
+import touchcare
```
dimagimon: F401 'touchcare' imported but unused

benrudolph:
- Merged in my branch here: https://github.com/dimagi/touchforms/pull/111 so I could more easily test functionality.
- Dropped use of `redis` python library and am using `Jedis` java library since I'll have to do way less configuration for it.

touchforms/backend/touchcare.py
```diff
@@ -248,25 +256,32 @@ def getIDsForValue(self, field_name, value):
...
-                        CaseDatabase(self.vars.get('host'), self.vars['domain'], self.vars['user_id'], 
-                                     self.auth, self.vars.get("additional_filters", {}),
-                                     self.vars.get("preload_cases", False)),
-                        False)
+            return CaseInstanceTreeElement(
+                instance.getBase(),
+                CaseDatabase(
+                    self.vars.get('host'),
+                    self.vars['domain'],
+                    self.auth,
```
snopoke: is `self.vars['user_id']` not required

czue:
@benrudolph seems like a cool change. based on https://github.com/dimagi/touchforms/commit/ea469e5592cfae9c2e8198383ab9c871826fd55d it looks like heartbeat was only ever enabled for offline cloudcare so I think we could just get rid if it entirely and then do the redis transition as a separate piece of work

czue:
(but no high-level concerns generally

touchforms/backend/touchcare.py
benrudolph: definitely should be, don't know how i missed that. thanks

benrudolph:
why do you say it's only enabled for offline usage? just went to check the new logs for touchforms. seems > 75% of the actions sent to touchforms are heartbeats. the reason i though redis was necessary is that i thought the memory footprint was going to get out of control if we didn't use it. (though as the commit you pointed to mentions, we can just change the staleness window)

czue:
My thinking was just that it was only added as part of that PR. Maybe it
still is needed (but seems fine for those sessions to fall out of memory
after a little while).

touchforms/backend/touchcare.py
benrudolph: oh sorry no i removed it on purpose. CaseDatabase was subclassing TouchfromsStorageUtility and needlessly adding the parameter: https://github.com/dimagi/touchforms/pull/118/files#diff-55395e6e3dc3202a4b20e5f392679013L170

snopoke: [?]

On 21 May 2015 at 19:12, Ben <notifications@github.com> wrote:

> In touchforms/backend/touchcare.py
> <https://github.com/dimagi/touchforms/pull/118#discussion_r30823591>:
>
> >          def from_bundle(inst):
> >              root = inst.getRoot()
> >              root.setParent(instance.getBase())
> >              return root
> >
> >          if 'casedb' in ref:
> > -            return CaseInstanceTreeElement(instance.getBase(),
> > -                        CaseDatabase(self.vars.get('host'), self.vars['domain'], self.vars['user_id'],
> > -                                     self.auth, self.vars.get("additional_filters", {}),
> > -                                     self.vars.get("preload_cases", False)),
> > -                        False)
> > +            return CaseInstanceTreeElement(
> > +                instance.getBase(),
> > +                CaseDatabase(
> > +                    self.vars.get('host'),
> > +                    self.vars['domain'],
> > +                    self.auth,
>
> oh sorry no i removed it on purpose. CaseDatabase was subclassing
> TouchfromsStorageUtility and needlessly adding the parameter:
> https://github.com/dimagi/touchforms/pull/118/files#diff-55395e6e3dc3202a4b20e5f392679013L170
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/dimagi/touchforms/pull/118/files#r30823591>.
>



-- 
Simon Kelly
Senior Engineer | Dimagi South Africa

benrudolph: This used to be the call to init:
```
def __init__(self, host, domain, user, auth, additional_filters=None, preload=False):		
       super(CaseDatabase, self).__init__(host, domain, auth, additional_filters, preload)
```
it completely ignores the user parameter and just calls the super method

snopoke: sorry, that was (thumbs up emoticon) from gmail :+1:

touchforms/backend/xformplayer.py
```diff
@@ -598,76 +594,85 @@ def open_form(form_spec, inst_spec=None, **kwargs):
...
 
 # fao mode only
-def submit_form(session_id, answers, prevalidated):
-    with global_state.get_session(session_id) as xfsess:
-        errors = dict(filter(lambda resp: resp[1]['status'] != 'success',
-                             ((_ix, xfsess.answer_question(answer, _ix)) for _ix, answer in answers.iteritems())))
+@modify_xform_session
+def submit_form(xform_session, answers, prevalidated):
+    errors = dict(filter(lambda resp: resp[1]['status'] != 'success',
+                         ((_ix, xform_session.answer_question(answer, _ix)) for _ix, answer in answers.iteritems())))
```
dimagimon: E501 line too long (117 > 115 characters)

touchforms/backend/xformplayer.py
```diff
@@ -598,76 +594,85 @@ def open_form(form_spec, inst_spec=None, **kwargs):
...
-    with global_state.get_session(session_id, override_state) as xfsess:
-        extra = {'lang': xfsess.get_lang()}
-        extra.update(init_context(xfsess))
-        return xfsess.response(extra, xfsess.cur_event)
+@modify_xform_session
+def set_locale(xform_session, lang):
+    ev = xform_session.set_locale(lang)
+    return xform_session.response({}, ev)
+
+@modify_xform_session
```
dimagimon: E302 expected 2 blank lines, found 1

touchforms/backend/touchcare.py
benrudolph: :+1:

touchforms/backend/xformplayer.py
```diff
@@ -613,81 +594,91 @@ def open_form(form_spec, inst_spec=None, **kwargs):
...
-    # just touch the session
-    with global_state.get_session(session_id) as xfsess:
-        return {}
+@modify_xform_session
+def current_question(xform_session, override_state=None):
+    extra = {'lang': xform_session.get_lang()}
+    extra.update(init_context(xform_session))
+    return xform_session.response(extra, xform_session.cur_event)
+
+def heartbeat():
```
dimagimon: E302 expected 2 blank lines, found 1

touchforms/backend/decorators.py
```diff
@@ -0,0 +1,13 @@
+from __future__ import with_statement
+
+
+def modify_xform_session(fn):
```
benrudolph: if anyone has a better name for this decorator i'd welcome it. couldn't think of anything better and i'm not super happy with it. basically it allows you to edit an xform thread safe and it'll cache the xform once you're done modifying it

benrudolph:
Ok feeling pretty good about this PR. will work to remove heartbeat like you mentioned @czue so we can get that quick win

touchforms/backend/decorators.py
```diff
@@ -0,0 +1,15 @@
+from __future__ import with_statement
+from functools import wraps
+
+
+def modify_xform_session(fn):
```
snopoke: Perhaps this could be called `require_xforms_session` since it injects the session into the function.

benrudolph: like it, thank you

benrudolph:
Ok, so this isn't working too well on staging. after way too much digging around in java code, i think it's because jedis doesn't use connection pooling? though that's kind of a guess. when i try to setup connection pooling, it doesn't work with jython because (my guess) one of the dependent jars uses some dynamic class lookups which is messing with it and causing a weird invalid argument error.

i like some of the other stuff in this PR though so i'm going to pull that out until i find a solution. the python redis library does connection pooling by default so that may be an option, but i'd have to figure out how to add that to class path in a neat way

snopoke:
What specifically isn't working well? I've got quite a bit of Java experience so perhaps could help figure out what's going on.

benrudolph:
@snopoke thanks, appreciate it. Here's what I know/have done. After deploying this to staging, things would work fine for around 10 minutes. Then redis would start failing and complain of a `Broken pipe`. I assumed this was related to connection pooling so I tried to implement [jedis's](https://github.com/xetorthio/jedis) `JedisPool` based on the (FAQ](https://github.com/xetorthio/jedis/wiki/Getting-started#using-jedis-in-a-multithreaded-environment). There's a line in there that says to define the `JedisPool` like so:
```
JedisPool pool = new JedisPool(new JedisPoolConfig(), "localhost");
```
When I did this initially I'd just get an error and the `JedisPool` was actually a `None` object. Took me a while to figure out I needed the [Apache Pool Jar](https://commons.apache.org/proper/commons-pool/). Once I included that, JedisPool was an actual object. However, when I tried to define it I'd get this error:
```
>>> pool = JedisPool(JedisPoolConfig(), "localhost")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
	at org.apache.commons.pool2.impl.BaseGenericObjectPool.setEvictionPolicyClassName(BaseGenericObjectPool.java:607)
	at org.apache.commons.pool2.impl.GenericObjectPool.setConfig(GenericObjectPool.java:321)
	at org.apache.commons.pool2.impl.GenericObjectPool.<init>(GenericObjectPool.java:117)
	at redis.clients.util.Pool.initPool(Pool.java:35)
	at redis.clients.util.Pool.<init>(Pool.java:22)
	at redis.clients.jedis.JedisPool.<init>(JedisPool.java:78)
	at redis.clients.jedis.JedisPool.<init>(JedisPool.java:13)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.lang.reflect.Constructor.newInstance(Constructor.java:408)
	at org.python.core.PyReflectedConstructor.constructProxy(PyReflectedConstructor.java:210)

java.lang.IllegalArgumentException: java.lang.IllegalArgumentException: Unable to create EvictionPolicy instance of type org.apache.commons.pool2.impl.DefaultEvictionPolicy
```
I did some source code diving, but nothing seemed too strange (also wasn't sure what I was looking for). This could very well be a weird Jython thing. Let me know if anything jumps out at you

snopoke:
Flip, I tried everything I could think of but no luck. It does seem that `Class.forName` is not well supported in Jython.

benrudolph:
No worries, thanks for taking a look

PR #122, opened 2015-05-23
Add filter_cases test

benrudolph:
@czue a little bit of a refactor, but should better test `filter_cases` so we'll know if it's broken. I want to modify `cases.js` in HQ so that it will handle 401 and 400 errors by showing the red box with the message. Then if we actually crash on a 500 we send a request to `/jserror/` so we receive an email. Also we'll now be able to see the actual filter expression when it fails on the xpath parsing which'll be super nice. (i'm wrestling with backbone not receiving the correct status code which is why i put the no merge label on)
elf: @snopoke

### DIFF
--- a/touchforms/backend/test_touchcare.py
+++ b/touchforms/backend/test_touchcare.py
@@ -78,7 +78,7 @@ class TouchcareTest(unittest.TestCase):
         else:
             self.fail()
 
-    def test_unauthenticated_query(self):
+    def test_unable_to_access_hq(self):
         try:
             touchcare.filter_cases(
                 '',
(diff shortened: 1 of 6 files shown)


benrudolph:
god for the guy who wrote every test in touchforms, i break them way too often. will fix

touchforms/backend/test_touchcare.py
```diff
@@ -0,0 +1,95 @@
...
+
+    def test_filter_cases(self):
+        filter_expr = "[case_name = 'rocky']"
+
+        resp = touchcare.filter_cases(
+            filter_expr,
+            {},
+            self.session_data,
+            form_context={
+                'cases': [self.case],
```
snopoke: Wouldn't it be better to have 2 cases here and make sure only one is returned?

snopoke:
Looks good to me. Test failure though.

PR #123, opened 2015-05-25
Test module

benrudolph:
Just opening this so I can run the tests on travis. been having trouble getting jython 2.7 to run on travis. also there's a bug in jython 2.5.3 that makes this not possible with that version. basically it can't find any of the test suites because `issubclass` fails to return correctly when testing if the suite is a subclass of TestCase
elf: @snopoke


touchforms/backend/tests/__init__.py
```diff
@@ -0,0 +1,4 @@
+from __future__ import absolute_import
+
+from .test_server import XFormServerTest
```
dimagimon: F401 'XFormServerTest' imported but unused

touchforms/backend/tests/__init__.py
```diff
@@ -0,0 +1,4 @@
...
+from .test_xformplayer import XFormPlayerTest
```
dimagimon: F401 'XFormPlayerTest' imported but unused

snopoke:
Seriously loathing jython right now

benrudolph:
tell me about it. killing me slowly. this probably should have been built in java

snopoke:
Still could be.....

PR #124, opened 2015-05-26
Push max date ten years into future

orangejenny:
http://manage.dimagi.com/default.asp?169219

There's similar code in submodules/touchforms-src/touchforms/formplayer/static/formplayer/script/deprecate/entrycontrols_type.js ... @benrudolph any idea if that directory is in use?

@TylerSheffels


benrudolph:
It's currently not in use. I tried to move everything that I thought wasn't being used into the deprecate folder until we are sure we can remove it

benrudolph:
:+1: will merge after tests

orangejenny:
@benrudolph Cool. Totally uninformed opinion, but is deleting files altogether any more likely to break things than renaming them?

benrudolph:
well, i updated the links that reference them so everything should still work. what i should do is log everytime they're loaded so I actually know if they're being used

orangejenny:
@benrudolph ah, that makes sense

PR #126, opened 2015-05-29
Do not allow empty submission

benrudolph:
http://manage.dimagi.com/default.asp?167990

@gcapalbo i'm totally convinced this is the expected behavior. if you have no required questions, why shouldn't you be able to submit an empty form? this seems like it should be pushed to the responsibility of the form maker. thought of this after i made the fix so happy discuss/merge

elf: @snopoke

### DIFF
--- a/touchforms/backend/test_global_state_manager.py
+++ b/touchforms/backend/test_global_state_manager.py
@@ -1,6 +1,7 @@
 import os
 import unittest
 import xformplayer
+from xcp import FormplayerSubmissionError
 
 CUR_DIR = os.path.dirname(__file__)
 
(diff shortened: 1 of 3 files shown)


touchforms/backend/test_global_state_manager.py
```diff
@@ -104,5 +105,34 @@ def test_instance_state(self):
...
+        xform_session = xformplayer.XFormSession(
+            self.xform,
+            instance=None,
+            **self.session_metadata
+        )
+        self.manager.cache_session(xform_session)
+
+        try:
+            xformplayer.submit_form(xform_session.uuid, {'0': None}, False)
+        except FormplayerSubmissionError, e:
```
dimagimon: F841 local variable 'e' is assigned to but never used

touchforms/backend/test_global_state_manager.py
```diff
@@ -104,5 +105,34 @@ def test_instance_state(self):
...
+    def test_form_submission(self):
+        xform_session = xformplayer.XFormSession(
+            self.xform,
+            instance=None,
+            **self.session_metadata
+        )
+        self.manager.cache_session(xform_session)
+        try:
+            xformplayer.submit_form(xform_session.uuid, {'0': 'Benjamin'}, False)
+        except FormplayerSubmissionError, e:
```
dimagimon: F841 local variable 'e' is assigned to but never used

benrudolph:
ah or do you mean to not allow a submission when there are no answers, i.e. the answer dictionary is empty. that would make more sense to me

snopoke:
looks ok code wise. will let @gcapalbo merge

gcapalbo:
@benrudolph The issue referenced in that ticket was that the form had all required questions, and yet somehow the user was able to submit while it still said "loading...", which caused the form to be submitted with empty values. For more context, see http://manage.dimagi.com/default.asp?167907.

benrudolph:
got it, makes sense

PR #129, opened 2015-06-05
sort ids inside CaseDB

czue:
http://manage.dimagi.com/default.asp?169413

@TylerSheffels

PR #130, opened 2015-06-05
sort ids inside CaseDB

czue:
version of https://github.com/dimagi/touchforms/pull/129 that hopefully fixes tests.

@dannyroberts cc @TylerSheffels

### DIFF
--- a/touchforms/backend/test_xformplayer.py
+++ b/touchforms/backend/test_xformplayer.py
@@ -75,7 +75,7 @@ class XFormPlayerTest(unittest.TestCase):
             self.xform,
             session_data=self.session_data,
             form_context={
-                'all_case_ids': ['legolas', 'gimli'],
+                'all_case_ids': ['legolas', 'sauron'],
                 'case_model': self.case
             }
         )
--- a/README.rst
+++ b/README.rst
@@ -89,3 +89,18 @@ Finally, compile the js files with ::
 Now if you run the grunt `watch` command, grunt will make sure to update compiled versions. For example, editting anything in the `fullform-ui` directory will automatically compile a new `fullform-ui.templates.js`. ::
 
     grunt watch
+
+
+Running Tests
+=============
+
+To run the backend tests you can just run the following ::
+
+     jython backend/test_server.py && jython backend/test_xformplayer.py && jython backend/test_touchcare.py && jython backend/test_global_state_manager.py
+
+To run tests you need to have Jython and `PhantomJS <http://phantomjs.org/>`_ installed.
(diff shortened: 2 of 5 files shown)


touchforms/backend/test_xformplayer.py
```diff
@@ -75,7 +75,7 @@ def mock_query(query_function, criteria):
             self.xform,
             session_data=self.session_data,
             form_context={
-                'all_case_ids': ['legolas', 'gimli'],
+                'all_case_ids': ['legolas', 'sauron'],
```
czue: this is pretty sketchy - basically as I can tell this line https://github.com/dimagi/commcare/blob/master/cases/src/org/commcare/cases/instance/CaseChildElement.java#L333 always tries to cache something with record ID of 0 when you initialize a form.

the sorting makes it so that the case we are checking (legolas) needs to come alphabetically before any others in the list of IDs.

this is bad, though no worse than the fact that if you flipped the order of `'legolas'` and `'gimli'` in the test on master it would also fail.

@benrudolph fyi

czue:
for future reference, another fun thing i learned was that you can throw a `java.lang.Throwable` inside your python code, and then catch it and call `t.printStackTrace()` to see where in the java it was being called

TylerSheffels:
build failed again

czue:
thanks, fixed. :white_check_mark:

touchforms/backend/test_xformplayer.py
benrudolph: hmm so if i understand correctly this worked before since we went in order through the `case_ids` and since we had the case for `legolas` in the `case_model` it stopped before checking gimli. now that we sort, gimli came first and broke it since the gimli case was not included in the `form_context` and thus looked for it on HQ which obviously isn't running during the tests. i think that that's fine, especially since this is really only used so we can test other functionality. it'll just make the tests kind of brittle. if we added the `gimli` case to the `form_context` i think it might've worked (https://github.com/dimagi/touchforms/blob/0efe0ccd3030b49d1ab248127fe2c4c7077ff6b5/touchforms/backend/test_touchcare.py#L67).

anyways thanks for fixing and diving into the java code. looked brutal

benrudolph:
:+1:  for docs and travis tweaks as well

touchforms/backend/test_xformplayer.py
czue: yeah, that's exactly right. I think the better fix would be to do the thing I suggest in the comment, and the even better fix would be to figure out why the mobile makes that call, but both of those seemed out of scope

PR #131, opened 2015-06-06
Add new exception for touchforms errors

gcapalbo:
@benrudolph @czue 
@emord code buddy

Added this so that we can inspect the actual response data when a touchforms error happens.

### DIFF
--- a/touchforms/formplayer/api.py
+++ b/touchforms/formplayer/api.py
@@ -41,15 +41,21 @@ class DigestAuth(TouchformsAuth):
     def to_dict(self):
         return {'type': self.type, 'key': self.key, 'username': self.username}
 
-            
-class TouchformsException(Exception):
+
+class TouchformsError(ValueError):
+
+    def __init__(self, *args, **kwargs):
+        self.response_data = kwargs.pop('response_data', {})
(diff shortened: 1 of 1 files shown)


touchforms/formplayer/api.py
```diff
@@ -41,16 +41,22 @@ def __init__(self, username, password):
     def to_dict(self):
         return {'type': self.type, 'key': self.key, 'username': self.username}
 
-            
-class TouchformsException(Exception):
-    pass
 
-class InvalidSessionIdException(TouchformsException):
+class TouchformsError(ValueError):
```
gcapalbo: Inherits from ValueError so that anything that's currently catching ValueError still works

touchforms/formplayer/api.py
```diff
@@ -41,16 +41,22 @@ def __init__(self, username, password):
     def to_dict(self):
         return {'type': self.type, 'key': self.key, 'username': self.username}
 
-            
-class TouchformsException(Exception):
```
gcapalbo: Nothing in the codebase catches the old `TouchformsException`, so replacing it with `TouchformsError`

benrudolph:
:+1:

PR #134, opened 2015-06-09
Use cached case list

benrudolph:
@czue @wspride this is the touchforms side, see: https://github.com/dimagi/commcare-hq/pull/7032 for more information

### DIFF
--- a/touchforms/backend/touchcare.py
+++ b/touchforms/backend/touchcare.py
@@ -26,7 +26,7 @@ from org.javarosa.core.model.instance import ExternalDataInstance
 from org.kxml2.io import KXmlParser
 
 from util import to_vect, to_jdate, to_hashtable, to_input_stream, query_factory
-from xcp import TouchFormsUnauthorized, TouchcareInvalidXPath
+from xcp import TouchFormsUnauthorized, TouchcareInvalidXPath, CaseNotFound
 
 logger = logging.getLogger('formplayer.touchcare')
 
@@ -233,7 +233,11 @@ class CaseDatabase(TouchformsStorageUtility):
 
         cases = self.cached_lookups[(field_name, value)]
         id_map = dict((v, k) for k, v in self.ids.iteritems())
-        return to_vect(id_map[c.getCaseId()] for c in cases)
+        try:
+            return to_vect(id_map[c.getCaseId()] for c in cases)
+        except KeyError:
+            # Case was not found in id_map
+            raise CaseNotFound
 
 
 class LedgerDatabase(TouchformsStorageUtility):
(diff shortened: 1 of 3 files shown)


touchforms/backend/touchcare.py
```diff
@@ -233,7 +233,11 @@ def getIDsForValue(self, field_name, value):
 
         cases = self.cached_lookups[(field_name, value)]
         id_map = dict((v, k) for k, v in self.ids.iteritems())
-        return to_vect(id_map[c.getCaseId()] for c in cases)
+        try:
+            return to_vect(id_map[c.getCaseId()] for c in cases)
+        except KeyError:
+            # Case was not found in id_map
```
benrudolph: probably superfluous

sheelio:
Is this going to be feature flagged or available for all CloudCare projects? (Happy to test on prod if feature flagged or staging if not).

benrudolph:
I was thinking about just deploying it since it seemed pretty small.
Feature flag would probably be the right way to go. We don't have that
functionality in touch forms so I'd only be able to add the flag in hq
which may work but might be a bit of a hack. This seemed kind of low risk
to me since the worst case scenario is stale data (famous last words).
@czue do you have any thoughts?

touchforms/backend/xformplayer.py
```diff
@@ -149,7 +150,13 @@ def load_form(xform, instance=None, extensions=None, session_data=None, api_auth
         preload_data=session_data.get('preloaders', {})
     )
 
-    form.initialize(instance is None, CCInstances(session_data, api_auth, form_context))
+    try:
+        session_data.get('additional_filters', {}).update({'use_cache': 'true'})
+        form.initialize(instance is None, CCInstances(session_data, api_auth, form_context))
+    except CaseNotFound:
```
czue: worth a comment explaining the logic behind this?

czue:
it doesn't seem obviously necessary to flag this to me. i think if we wanted to we could easily pass a flag from HQ to touchforms in the params though.

QA might make sense though - and if that's going to be complicated on staging then switching to a flag might still be the best plan.m

benrudolph:
sounds good. i was just about to write that it's probably best to throw it up on staging first

touchforms/backend/xformplayer.py
benrudolph: definitely reasonsable

touchforms/backend/xformplayer.py
```diff
@@ -149,7 +150,17 @@ def load_form(xform, instance=None, extensions=None, session_data=None, api_auth
...
+        # Touchforms repeatedly makes a call to HQ to get all the case ids in its universe. We can optimize
+        # this by caching that call to HQ. However, when someone adds a case to that case list, we want to make sure
```
dimagimon: E501 line too long (116 > 115 characters)

benrudolph:
self merging, see comment on this PR: https://github.com/dimagi/commcare-hq/pull/7032. don't want to deploy without the second half in

czue:
:+1: thanks

PR #135, opened 2015-06-12
Add doc string for storage utility

benrudolph:
This thing has been a bit of a enigma for me, so I'm trying to clear it up for folks in the future. @czue we do need that call to HQ to grab all the case ids because we do not know whether or not the form will reference one of those cases, in say, a count of all the cases.

elf: @dannyroberts

### DIFF
--- a/touchforms/backend/touchcare.py
+++ b/touchforms/backend/touchcare.py
@@ -105,6 +105,16 @@ class StaticIterator(IStorageIterator):
 
 
 class TouchformsStorageUtility(IStorageUtilityIndexed):
+    """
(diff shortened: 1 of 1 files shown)


touchforms/backend/touchcare.py
```diff
@@ -105,6 +105,16 @@ def nextID(self):
 
 
 class TouchformsStorageUtility(IStorageUtilityIndexed):
+    """
+    The TouchformsStorageUtility provides an interface for working with the case database. The mobile phone
```
dimagimon: W291 trailing whitespace

czue:
thanks for docs!

PR #138, opened 2015-06-18
Revert "Revert "Use classes instead of ids""

benrudolph:
Reverts dimagi/touchforms#137

Going to look into this...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,4 +9,5 @@ branches:
       - master
 install:
   - "pip install -r touchforms/test_requirements.txt"
+  - "./install_dependencies.sh"
 script: "./run_tests.sh"
--- a/install_dependencies.sh
+++ b/install_dependencies.sh
@@ -1,0 +1,2 @@
+#!/bin/sh
+cd touchforms && npm install && node_modules/.bin/grunt build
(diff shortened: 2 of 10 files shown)


benrudolph:
@czue Ok, so I think I know what went wrong with this. I can repeat the same error in the ticket if the new templates are not compiled. My guess is, that there was lag time when the code was updated but the new templates hadn't been generated. do you think that's possible? either that or the old templates file was cached and we aren't properly breaking the cache.

it may be time to rewrite all this code with knockout so we can use knockout templates.

benrudolph:
also explains why i wasn't able to reproduce even though i was logged in as him

benrudolph:
the hard part about going to knockout is that it will require quite a big paradigm shift in the JS code

benrudolph:
dove into this for a few hours. switching to knockout is probably going to require refactoring nearly everything

czue:
my gut is that rewriting all of the javascript is not worth doing now given other more pressing scale stuff. let me know if you disagree with that.

czue:
also not sure one how the template caching works so not sure about your initial question - might need a bit more context to find out / answer. feel free to grab me sometime tomorrow and/or write up a few more details about how you think this was breaking

benrudolph:
yea fair, definitely gonna focus on the scale tests. got deep into a refactor which was actually pretty fun to do

PR #140, opened 2015-07-01
don't fail hard on edge-case error, better error logging

czue:
http://manage.dimagi.com/default.asp?170392 for context

@dannyroberts @TylerSheffels or whoever


touchforms/formplayer/static/formplayer/script/xforgasm.js
```diff
@@ -266,6 +266,7 @@ function WebFormSession(params) {
             try {
                 callback(resp);
             } catch (err) {
+                console.error(err);
```
snopoke: console log

czue: (this was intentional)

PR #141, opened 2015-07-02
add static date function

czue:


touchforms/backend/handlers/static.py
```diff
@@ -32,3 +32,38 @@ def getName(self):
...
+    - Datetimes with microseconds (UTC): 2015-07-01T14:52:58.473000Z
+
+    Fails hard if not passed a valid non-empty string in one of these formats
+    """
+
+    @classmethod
+    def slug(self):
+        return 'static-date'
+
+    def __init__(self, name, value):
```
czue: parsing logic is pretty brittle but i think that's ok

czue:
@wpride 

cc @TylerSheffels @dannyroberts

PR #146, opened 2015-07-22
don't fail if unable to find media map for a question

czue:
http://manage.dimagi.com/default.asp?176097

### DIFF
--- a/touchforms/formplayer/static/formplayer/script/fullform-ui.js
+++ b/touchforms/formplayer/static/formplayer/script/fullform-ui.js
@@ -394,22 +394,31 @@ function Question(json, parent) {
       //this.control.restore_ui_state(uistate);
     }
 
-    var add_multimedia = function(attrib, control) {
-      if (self.hasOwnProperty(attrib) && self[attrib]) {
-        var mediaSrc = getForm(self).adapter.render_context.resourceMap(self[attrib]);
-        if (mediaSrc) {
-          control.attr("src", mediaSrc);
-          $mediaContainer = $('<div>');
-          $mediaContainer.append(control);
-          var $widget = self.$container.find('#widget');
-          if ($widget.length) {
-            $widget.append($mediaContainer);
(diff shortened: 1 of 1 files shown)


czue:
the two-space tabs hurt my eyes, but i spent more time fiddling with whitespace than i did implementing the fix before bailing

czue:
@emord buddy

touchforms/formplayer/static/formplayer/script/fullform-ui.js
```diff
@@ -394,22 +394,31 @@ function Question(json, parent) {
...
-          var $widget = self.$container.find('#widget');
-          if ($widget.length) {
-            $widget.append($mediaContainer);
-          } else {
-            $capt.append($mediaContainer);
+    var add_multimedia = function (attrib, control) {
+      var getMediaSrc = function (attrib) {
+        if (self.hasOwnProperty(attrib) && self[attrib]) {
+          try {
+            return getForm(self).adapter.render_context.resourceMap(self[attrib]);
```
benrudolph: this kind of seems like a hammer. if anything were to break in the resourceMap function we'd never know. kind of like using `except Exception`. shouldn't the change be made in the `resourceMap`? https://github.com/dimagi/commcare-hq/blob/fad41eb69d050c6e773db125fe25b85e192b2469/corehq/apps/cloudcare/static/cloudcare/js/backbone/apps.js#L703

benrudolph:
haha that sounds like `fullform-ui.js`!

touchforms/formplayer/static/formplayer/script/fullform-ui.js
benrudolph: maybe as a halfway, we could at least log the error in the catch function `console.error(err)`

czue: legit. updated

PR #150, opened 2015-07-26
log session ids

czue:


czue:
@benrudolph @wpride

benrudolph:
Looks good thanks for adding

PR #151, opened 2015-07-29
Locking debug

czue:
@benrudolph just adding logging and reducing purge frequency

related to http://manage.dimagi.com/default.asp?175484


touchforms/backend/xformplayer.py
```diff
@@ -38,12 +38,10 @@
 from xcp import CaseNotFound
 import persistence
 import settings
-import logging
```
benrudolph: oops good catch

benrudolph:
:+1:

PR #152, opened 2015-08-05
More lock cleanup

czue:
I think this is the final issue with http://manage.dimagi.com/default.asp?175484

For some reasons the sessions were all sharing a global lock. this just removes that and has them use the session-specific locks.

I'm not sure whether we think we should QA this extensively or not. I filled in some forms etc. and didn't see any issues, and was very careful about searching for usages of both lock types. Open to feedback/pushback on that.

@benrudolph @snopoke

### DIFF
--- a/touchforms/backend/xformplayer.py
+++ b/touchforms/backend/xformplayer.py
@@ -170,7 +170,6 @@ class SequencingException(Exception):
 class XFormSession(object):
     def __init__(self, xform, instance=None, **params):
         self.uuid = params.get('uuid', uuid.uuid4().hex)
-        self.lock = threading.Lock()
         self.nav_mode = params.get('nav_mode', 'prompt')
         self.seq_id = params.get('seq_id', 0)
 
--- a/touchforms/backend/xformplayer.py
+++ b/touchforms/backend/xformplayer.py
@@ -718,9 +718,17 @@ def current_question(xform_session, override_state=None):
 
 
 def heartbeat(session_id):
-    # just touch the session
-    with global_state.get_session(session_id) as xfsess:
-        return {}
+    """
+    The heartbeat will update the session's last seen timestamp and sequence ID.
+    """
+    # todo: cz as of 8/5/2015 I'm pretty sure this isn't necessary. persistence can
+    # just pop this back into the context whenever necessary so no longer super relevant
+    # to keep around in memory while the tab is open.
+    with global_state.get_lock(session_id):
+        # this updates the session stuff and persists it
+        with global_state.get_session(session_id):
+            {}
+
 
 def next_event (xfsess):
     ev = xfsess.next_event()
--- a/touchforms/backend/xformplayer.py
+++ b/touchforms/backend/xformplayer.py
@@ -207,24 +207,24 @@ class XFormSession(object):
         }
         self.update_last_activity()
 
+    def _assert_locked(self):
+        if not global_state.get_lock(self.uuid).locked():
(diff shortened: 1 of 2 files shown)


czue:
(going to test locally a bit more)

czue:
after further testing and discussion with @snopoke I don't actually think this is going to solve anything, but is hopefully still an improvement

benrudolph:
yea i'm pretty perplexed on the logs vs. the code. the logs do seem to point to a global lock being grabbed, but i think we'd see that in the logs.

from what i understand, this lock was totally superfluous cause the session should be locked anytime you were modifying the XFormSession object so the internal lock on the object actually did nothing. also each request would have a new XFormSession object and thus a different lock. is that why you think it won't make a difference?

this definitely makes the code clearer!

touchforms/backend/xformplayer.py
```diff
@@ -718,9 +720,17 @@ def current_question(xform_session, override_state=None):
...
-        return {}
+    """
+    The heartbeat will update the session's last seen timestamp and sequence ID.
+    """
+    # todo: cz as of 8/5/2015 I'm pretty sure this isn't necessary. persistence can
+    # just pop this back into the context whenever necessary so no longer super relevant
+    # to keep around in memory while the tab is open.
+    with global_state.get_lock(session_id):
+        # this updates the session sequence ID and timestamp and persists it
+        with global_state.get_session(session_id):
```
benrudolph: we can totally get rid of this function as far as i know. we only make a request to `heartbeat` via the `restarter.sh` and that call doesn't include a session id so it returns a 400 response and this code never executes. ideally, this function should just return a message stating the server is up and a response 200

czue:
> from what i understand, this lock was totally superfluous cause the session should be locked anytime you were modifying the XFormSession object so the internal lock on the object actually did nothing. also each request would have a new XFormSession object and thus a different lock. is that why you think it won't make a difference?

you didn't actually have the lock. you only got the lock to pull the session out of the hash, but once `get_session` returned the lock was released. then the session specific lock kicked in with the second `with` statement. so this at least reduces two locks to a single lock, but still doesn't answer the other question.

I'm not 100% convinced the latest issues have nothing to do with lock contention and more to do with some other resource on the machine getting jammed up.

touchforms/backend/xformplayer.py
czue: does the front end not call it anymore?

benrudolph: nope: https://github.com/dimagi/touchforms/pull/119

czue: but why are the logs still logging heartbeats per sessions?

benrudolph: hmm, these are the only two lines i see with `heartbeat` in them in your log [doc](https://docs.google.com/document/d/1IwQCOaqZA9ptUmc7RrECzr1zjzPgYfAbvbC52lxAo4A/edit)

```
INFO 2015-07-30 12:28:01,858 xformserver 3093 8687 Received action heartbeat for session <unknown>
INFO 2015-07-30 12:28:01,858 xformserver 3093 8687 Finished processing action heartbeat in 0.999927520752 ms for session <unknown> in domain '<unknown>'
```

czue: huh, you're right, they're all `<unknown>`

czue: https://github.com/dimagi/touchforms/pull/153

PR #153, opened 2015-08-05
remove heartbeat

czue:
@benrudolph

### DIFF
--- a/touchforms/backend/xformserver.py
+++ b/touchforms/backend/xformserver.py
@@ -223,8 +223,7 @@ def handle_request(content, server):
             return xformplayer.current_question(content['session-id'], override_state=override_state)
 
         elif action == xformplayer.Actions.HEARTBEAT:
-            return xformplayer.heartbeat(content['session-id'])
-
+            return {}
         elif action == xformplayer.Actions.EDIT_REPEAT:
             ensure_required_params(['session-id', 'ix'], action, content)
             return xformplayer.edit_repeat(content['session-id'], content['ix'])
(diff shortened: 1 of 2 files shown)


benrudolph:
cool. this'll require some coordination with the deployer since this will cause the `restarter.sh` to endlessly restart touchforms. happy to do it once i get back on a reliable internet connection or you can self-merge and make sure the restarter.sh gets updated when the deploy takes place

czue:
(don't merge - I'll look at coordination tomorrow)

czue:
er - i'll look at coordination when i get a chance (might not be today)

benrudolph:
i can do this today, i'm deploying so it'll probably be easiest

czue:
@benrudolph i was thinking about this more - can you explain why this requires any coordination? the api still exists, it's just the function that doesn't do anything.

czue:
(leaving label on but if that sounds right to you feel free to merge)

touchforms/backend/xformserver.py
```diff
@@ -223,8 +223,7 @@ def handle_request(content, server):
             return xformplayer.current_question(content['session-id'], override_state=override_state)
 
         elif action == xformplayer.Actions.HEARTBEAT:
-            return xformplayer.heartbeat(content['session-id'])
```
benrudolph: @czue, before, the restarter.sh would hit this route and error because there was no session-id. now that we don't check for the session-id it will return a 200 status code. before we were checking for a 500 (or 400 i forget). definitely better to check for a 200 though so i'm all for this change

czue:
makes sense, thanks

czue:
thanks for dealing with this!

PR #165, opened 2015-09-15
Add auto sync back in

benrudolph:
I think this just got mixed up in a merge. Copied from this PR: https://github.com/dimagi/commcare-hq/pull/7561/files

Basically since the checkbox to autosync never rendered, it was always false and thus the debug tool never loaded

http://manage.dimagi.com/default.asp?182140#1012480

@czue


czue:
:+1: when tests pass

PR #166, opened 2015-09-15
specify unsubscribes

benrudolph:
@czue fixes the dirty flag when navigating away from the page: http://manage.dimagi.com/default.asp?182149. (has test coverage). still doesn't fix when navigating away from form to another form

### DIFF
--- a/touchforms/formplayer/static/formplayer/script/formnav-all.js
+++ b/touchforms/formplayer/static/formplayer/script/formnav-all.js
@@ -10,7 +10,13 @@ function xformAjaxAdapter (formSpec, sessionData, savedInstance, ajaxfunc, submi
   this.render_context = render_context;
   this.answerCallback = answerCallback;
 
-  $.unsubscribe('formplayer');
+  $.unsubscribe([
+      'formplayer.submit-form',
+      'formplayer.delete-repeat',
+      'formplayer.new-repeat',
+      'formplayer.answer-question',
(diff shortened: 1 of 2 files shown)


touchforms/formplayer/static/formplayer/script/formnav-all.js
```diff
@@ -10,7 +10,13 @@ function xformAjaxAdapter (formSpec, sessionData, savedInstance, ajaxfunc, submi
...
   this.answerCallback = answerCallback;
 
-  $.unsubscribe('formplayer');
+  $.unsubscribe([
+      'formplayer.submit-form',
+      'formplayer.delete-repeat',
+      'formplayer.new-repeat',
+      'formplayer.answer-question',
+      'formplayer.evaluate-xpath'
+  ].join(' '));
```
benrudolph: since the event model is global, it would unsubscribe the `formplayer.dirty` event.

touchforms/formplayer/static/formplayer/script/fullform-ui.js
```diff
@@ -318,7 +318,7 @@ function Question(json, parent) {
     self.afterRender = function() { self.entry.afterRender(); };
 
     self.onchange = _.throttle(function() {
-        $.publish('formplayer.dirty', true);
+        $.publish('formplayer.dirty');
```
benrudolph: just a typo, makes no difference but slightly more clear

PR #171, opened 2015-10-07
Update datepicker format commenting

orangejenny:
@benrudolph does this seem like a reasonable change? i maintain that jquery using y and yy instead of yy and yyyy is a bad idea that's going to confuse many people reading this code.


benrudolph:
:+1:

PR #172, opened 2015-10-07
don't set date on no answer

benrudolph:
@orangejenny cc: @TylerSheffels


benrudolph:
given this has caused 3 bugs, i think it's almost time for me to put infrastructure to test html...

orangejenny:
looks good. merging, given lack of test infrastructure.

PR #173, opened 2015-10-07
Use unwrapObservable for check

benrudolph:
@czue, this should have a test, but i need to take a half day to really understand how form indexing is working. the nomenclature is still a bit foggy to me, i.e. `-0,1:2` still makes no sense. i think it'd probably be worth the time investment because from the people i've talked to no one really knows how this works that well

cc: @TylerSheffels

### DIFF
--- a/touchforms/formplayer/static/formplayer/script/fullform-ui.js
+++ b/touchforms/formplayer/static/formplayer/script/fullform-ui.js
@@ -32,7 +32,7 @@ function getIx(o) {
     var ix = o.rel_ix();
     while (ix[0] == '-') {
         o = o.parent;
-        if (!o || o.rel_ix() === undefined) {
+        if (!o || ko.utils.unwrapObservable(o.rel_ix) === undefined) {
             break;
         }
         if (o.rel_ix().split(',').slice(-1)[0].indexOf(':') != -1) {


czue:
Nice. Does this fix http://manage.dimagi.com/default.asp?183796?

TylerSheffels:
:+1:

benrudolph:
yep

czue:
agree that no one understands how this works and that that's a bad thing. the only question I would have about how much to get into it would be depending on how much of this code the changes that @wpride has planned will get rid of. my understanding is that it might be all of it, but I could be wrong.

benrudolph:
i think for the sqlite infrastructure we're planning on keeping the same api to the frontend and i'm pretty sure this will all stay unchanged, but yea will to confirm

czue:
there's a second refactor (targeting march) that plans to move all form and suite navigation to the mobile apis

benrudolph:
oh interesting. i'd imagine that'd have a lot to do with the backbone portion on hq as well. would like to be interested to hear how it's going to be implemented

wpride:
Yeah, I think ~all of this will be unused in the near future, with this
logic either in Enketo or the Java libs. I'm overdue to dump my current
thoughts about the refactor

benrudolph:
holy shit

benrudolph:
that'd be awesome. i'm still a bit confused on how we'd escape from rendering the entry controls ourselves. you're planning on using the enketo lib to do the display?

wpride:
Yeah that's what I'm hoping, though I don't want to sound overly confident
about the Enketo since I've just started playing with it and I'm still
discovering new incompatibilities/discrepancies between our formats (for
example they don't have setvalues which we make quite heavy use of)

The way I'm currently thinking about it:
You make an app/login/menu/case/form selection using an API-ish format that
the CLI uses
We retrieve the XForm
We replace all the external instance refs with the parallel XML (will get
tricky/impossible with ledgers and larger caseDBs for sure without some
clever optimizations)

-> We'd probably need to set it up so that if you try to hit a case or
ledger you don't have loaded you make a server request
-> Eventually we'd probably want to use browser storage

We fire ALL of the setvalue event=ready BEFORE sending the XForm to enketo
(so all the loaded case properties are just there as hard coded XML)
Put XForm through Enketo Transformer
Send instance and form to Enketo Core
Enketo Core does the form entry
We receive the "completed" form and do some post processing (firing all
setvalue event=revalidates, basically)
We submit the completed form to server

PR #174, opened 2015-10-13
Log domain for TF requests

benrudolph:
finally looked into why touchcare-filter-cases wasn't getting its domain logged. turns out we use `session_data` instead of `session-data`. nice piece of tech debt. put it in my back log to fix, but for now it's more important to get to the bottom of these 539 second requests


benrudolph:
@czue @TylerSheffels

orangejenny:
>  turns out we use session_data instead of session-data

aww, that's rough

czue:
that's hilariously terrible.

PR #177, opened 2015-10-28
Fix firefox img src

benrudolph:
@esoergel this ensures that broken images don't show up on FF


proteusvacuum:
:fire: :rabbit: :smiley_cat:

esoergel:
Nice!  @benrudolph fighting the good fight!

benrudolph:
:facepunch: :facepunch:

PR #179, opened 2015-11-14
Better log format

benrudolph:
@snopoke @czue

### DIFF
--- a/touchforms/backend/xformserver.py
+++ b/touchforms/backend/xformserver.py
@@ -165,7 +165,7 @@ def handle_request(content, server):
         session_id = content['session-id']
 
     action = content['action']
-    logger.info('Received action %s for session %s' % (action, session_id))
+    logger.info('event=received action=%s session=%s' % (action, session_id))
     nav_mode = content.get('nav', 'prompt')
     try:
         # Formplayer routes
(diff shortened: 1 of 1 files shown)


touchforms/backend/xformserver.py
```diff
@@ -165,7 +165,7 @@ def handle_request(content, server):
         session_id = content['session-id']
 
     action = content['action']
-    logger.info('Received action %s for session %s' % (action, session_id))
+    logger.info('event=received action=%s session=%s' % (action, session_id))
```
czue: wondered whether we should remove spaces from the events, but assume you have a plan for the parser

benrudolph: what do you mean by removing the spaces from the events? i went half way to implement the [datadog canonical log format](http://docs.datadoghq.com/guides/logs/) but decided i'd probably need to make a new logger for that since it uses unix timestamps for dates which isn't very useful when you're looking at logs. i figured this would be easy enough to parse `split()` and then split on the `=` sign

czue: sorry i'm an idiot - read it as "event=received action"

snopoke: Just a general comment but often the way to do this kind of logging is to add the parameters to the 'extra' arg which allows you to configure the output format in the logging config.
```
details = {'action': action, 'session': session_id}
logger.info('reveived %s action for session %s', action, sessino_id, extra=details)
```

I would also have thought that a format that is just space separated keys would be easiest to parse:
```
{date} {session} {domain} {received / processed} {action}
```

benrudolph: good to know. my logging knowledge is pretty elementary. i'll play around with that

PR #181, opened 2015-11-17
Better log format

benrudolph:
@czue @snopoke take 2. this'll obviate the need for a parser and it seems useful to have a datadog logger around in the future for touchforms in case we want to do more analytics

### DIFF
--- a/touchforms/backend/settings.py
+++ b/touchforms/backend/settings.py
@@ -41,7 +41,7 @@ POSTGRES_DATABASE = {
 
 ### LOGGING VARIABLES ###
 FORMPLAYER_LOG_FILE = 'formplayer-dev.log'
-DATADOG_FORMPLAYER_LOG_FILE = 'datadog-dev.log'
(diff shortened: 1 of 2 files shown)


touchforms/backend/settings.py
```diff
@@ -41,9 +41,11 @@
 
 ### LOGGING VARIABLES ###
 FORMPLAYER_LOG_FILE = 'formplayer-dev.log'
+DATADOG_FORMPLAYER_LOG_FILE = 'datadog-dev.log'
 
 formats = {
     'verbose': '%(levelname)s %(asctime)s %(module)s %(process)d %(thread)d %(message)s',
+    'datadog': 'touchforms.%(metric)s %(timestamp)s %(value)s metric_type=%(metric_type)s %(message)s'
```
benrudolph: tried to make `timestamp` work with `asctime` but couldn't get it

benrudolph:
https://github.com/dimagi/commcarehq-ansible/pull/264 corresponding pr. also cc: @emord

snopoke:
nice

PR #188, opened 2015-12-08
have caption have normal font-weight

benrudolph:
@orangejenny this overrides our default bolding of label text (not in love with that, but suppose it's nice to have the markdown bold work as expected). the other thing is a bit trickier. since markdown adds a `<p>` block, it messes with the required asterix. the only thing i can think of doing is floating it to the left like in this screenshot:
screen shot 2015-12-08 at 3 59 58 pm
i don't like it too much though. do you have any ideas?

cc: @czue

### DIFF
--- a/touchforms/formplayer/static/formplayer/style/webforms.css
+++ b/touchforms/formplayer/static/formplayer/style/webforms.css
@@ -93,11 +93,13 @@ input:invalid {
 }
 
 
-.required .caption:before {
+.required .caption::before {
   content: '*';
   font-weight: bold;
   color: #c0392b;
   margin: 0 3px;
+  right: 0;
+  position: absolute;
 }
 
 .caption {
--- a/touchforms/formplayer/static/formplayer/style/webforms.css
+++ b/touchforms/formplayer/static/formplayer/style/webforms.css
@@ -100,6 +100,10 @@ input:invalid {
   margin: 0 3px;
 }
 
+.caption {
+  font-weight: normal;
+}
+
 .dirty .widget {
     background-color: lightblue;
 }


orangejenny:
Looks pretty reasonable. Using both `float` and `position` in `.caption` strikes me as weird, even though you're not using the position except to establish it as a positioned parent.

One different thought: what about placing the asterisk on the right instead of the left? You'd be able to scan a form for required questions more easily, and the CSS is simple (only adjustment to the current production styles is adding `positon: absolute; right: 0;` to `.required .caption::before`).

right

benrudolph:
i'm game for the right hand side, i tested it and thought it'd be shot down, but i do like the simplicity of the code. (yes position is just so the pseudo element's position absolute works properly)

benrudolph:
@dimagi/product any objections to moving the asterisk to the right hand side?

amsagoff:
Not at all, think it's a positive change!

benrudolph:
aight g2g @orangejenny once tests pass

orangejenny:
looks good

touchforms/formplayer/static/formplayer/style/webforms.css
```diff
@@ -93,11 +93,17 @@ input:invalid {
 }
 
 
-.required .caption:before {
+.required .caption::before {
```
orangejenny: do these both work? i've only seen `::` but obviously this was working before with just `:`.

benrudolph: i believe `:` is css2 and `::` is 3

benrudolph: for our browsers it doesn't make a difference

orangejenny: cool, i learned something

PR #187, opened 2015-12-05
Cleanup cloudcare

benrudolph:
@czue @dimagi/js-team and the touchforms portion. unfortunately the only real way to read this is commit-by-commit. had to do it very piecemeal

screen shot 2015-12-06 at 12 29 16 am and i added tests!

PR #189, opened 2015-12-10
ensure that resp has responseJSON

benrudolph:
@czue


benrudolph:
this more closely resembles what happens here: https://github.com/dimagi/touchforms/pull/187/files#diff-6648044907326d9b3a5c65869521170fL203. should help get to the root cause for this: http://manage.dimagi.com/default.asp?189055#1058180

PR #192, opened 2015-12-17
Debugger tweaks

benrudolph:
This does a couple things based on feedback i got from some of the SA office. @wpride (since zue is intermittent and @proteusvacuum you showed a bit of interest, i tag you as the hq dev!) 

- allows you to click the entire blue bar to expand the debugger (instead of the tiny little arrows)
- resizes the debugger box to the full viewport
- eliminates the weird xpath result box and puts the query box on top (and much larger since the expressions tend to be bigger than the result)

screenshot:
screen shot 2015-12-17 at 10 14 14 pm

buddy: @czue


proteusvacuum:
:+1: :cloud:

PR #197, opened 2016-01-19
fix obscure json parsing issue

czue:
http://manage.dimagi.com/default.asp?190983

This is pretty weird. JQuery is raising a parsererrror even though the response is valid json. What's even weirder is that returning the same response to a different request doesn't fail - and returning a trivial response to the same request does fail. 

So I think it must be some weird bug/edge case with a super-complicated request triggering a deep bug in jquery.

Others have run into this and one simple workaround is to just bypass jquery's automatic json parser, which seems to work fine http://stackoverflow.com/questions/5095307/jquery-json-response-always-triggers-a-parseerror

Not a very satisfying answer, but seems to fix the issue.

@benrudolph cc @esoergel

### DIFF
--- a/touchforms/formplayer/static/formplayer/script/webformsession.js
+++ b/touchforms/formplayer/static/formplayer/script/webformsession.js
@@ -137,15 +137,14 @@ WebFormSession.prototype.serverRequest = function (requestParams, callback, bloc
 
     this.numPendingRequests++;
     this.onLoading();
-
     $.ajax({
             type: 'POST',
             url: url,
             data: JSON.stringify(requestParams),
-            dataType: "json",
+            dataType: "text"  // we don't use JSON because of a weird bug: http://manage.dimagi.com/default.asp?190983
         })
-        .success(function(resp) { self.handleSuccess(resp, callback); })
-        .fail(self.handleFailure.bind(self));
+        .success(function(resp) { self.handleSuccess(JSON.parse(resp), callback); })
+        .fail(function (resp, textStatus) { self.handleFailure(JSON.parse(resp), textStatus); });
 };
 
 /*
--- a/touchforms/formplayer/views.py
+++ b/touchforms/formplayer/views.py
@@ -228,14 +228,14 @@ def player_proxy(request):
                                  content_type="text/json", auth=DjangoAuth(auth_cookie))
 
         _track_session(request, json.loads(data), json.loads(response))
-        return HttpResponse(response)
+        return HttpResponse(response, content_type='application/json')
     except IOError:
         logging.exception('Unable to connect to touchforms.')
         msg = _(
             'An error occurred while trying to connect to the CloudCare service. '
             'If you have problems filling in the rest of your form please report an issue.'
         )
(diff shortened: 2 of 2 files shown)


touchforms/formplayer/static/formplayer/script/webformsession.js
```diff
@@ -155,6 +154,7 @@ WebFormSession.prototype.serverRequest = function (requestParams, callback, bloc
  */
 WebFormSession.prototype.handleSuccess = function(resp, callback) {
     var self = this;
+    var resp = JSON.parse(resp);
```
dimagimon: 'resp' is already defined.

touchforms/formplayer/views.py
```diff
@@ -228,14 +228,14 @@ def player_proxy(request):
                                  content_type="text/json", auth=DjangoAuth(auth_cookie))
 
         _track_session(request, json.loads(data), json.loads(response))
-        return HttpResponse(response)
+        return HttpResponse(response, content_type='application/json')
```
czue: this didn't fix it, even though [some people suggested it might](http://stackoverflow.com/a/249758/8207). Still seems like a good thing to do.

benrudolph:
wow that is bizarro

benrudolph:
error is most likely due to the fact that we are passing in json instead of text to the `handleSuccess` function here: https://github.com/dimagi/touchforms/blob/master/touchforms/spec/javascripts/webformsession_spec.js#L177. sorry about the lack of docs on how to run the frontend tests, will update that

esoergel:
Build failure looks real: https://travis-ci.org/dimagi/touchforms/builds/103329742
```
34 specs, 2 failed

WebForm WebFormSession Should handle error in callback
  SyntaxError: Unable to parse JSON string in http://localhost:8889/__src__/script/webformsession.js (line 157)
  SyntaxError: Unable to parse JSON string
    at http://localhost:8889/__src__/script/webformsession.js:157
    at http://localhost:8889/__spec__/webformsession_spec.js:168
WebForm WebFormSession Should handle error in response
  SyntaxError: Unable to parse JSON string in http://localhost:8889/__src__/script/webformsession.js (line 157)
  SyntaxError: Unable to parse JSON string
    at http://localhost:8889/__src__/script/webformsession.js:157
    at http://localhost:8889/__spec__/webformsession_spec.js:177
```

czue:
thanks looking at the build now

touchforms/formplayer/static/formplayer/script/webformsession.js
```diff
@@ -155,6 +154,7 @@ WebFormSession.prototype.serverRequest = function (requestParams, callback, bloc
  */
 WebFormSession.prototype.handleSuccess = function(resp, callback) {
     var self = this;
+    resp = JSON.parse(resp);
```
benrudolph: actually this would also need to go into the `handleFailure` i think

czue:
alrighty, cleaned this up a bit and ran tests locally this time so should be good to go (but fine to wait)

benrudolph:
:boom:

PR #201, opened 2016-02-23
fallback to old behavior if 'default' not present

wpride:
@gcapalbo


gcapalbo:
ran sms tests locally, and they pass - just waiting on travis

PR #205, opened 2016-03-09
Fix test

wpride:
status filter tests are no longer meaningful now that we added this filtering hack https://github.com/dimagi/touchforms/commit/c7ac15cb7cbad479a72a1ebae6dec3f98988dadf

fixes failing test


benrudolph:
:+1:

PR #206, opened 2016-03-10
use proper timezoning for java dates

wpride:
Adding don't merge because I'm not convinced there isn't a better way to fix this/not clear why this broke yet

see http://manage.dimagi.com/default.asp?212637 

Basically in the XML instance we store the timezone as

`<n0:timeEnd>2016-03-10T15:28:02.589+03</n0:timeEnd>`

Then we load this into the context as:

Time End:  2016-03-10 12:28:02.589000

(IE convert to UTC)

But THEN we're calling this server time (locally for me EAT) in the static handler:

Setting value to parsed:  Thu Mar 10 00:28:02 EAT 2016  from value  2016-03-10T12:28:02Z

And then returning:

Returning ret: Thu Mar 10 00:28:02 EAT 2016 class java.util.Date

### DIFF
--- a/touchforms/backend/handlers/static.py
+++ b/touchforms/backend/handlers/static.py
@@ -1,6 +1,6 @@
 from customhandlers import TouchformsFunctionHandler
 from java.text import SimpleDateFormat
-
+from java.util import TimeZone
 class StaticFunctionHandler(TouchformsFunctionHandler):
     """
     A function handler that lets you register a static value associated with a function.
@@ -65,5 +65,7 @@ class StaticDateFunctionHandler(StaticFunctionHandler):
                 # remove microseconds if necessary
                 if len(value) == 27:
                     value = '%sZ' % value[:19]
-                parsed_value = SimpleDateFormat("yyyy-MM-dd'T'hh:mm:ss'Z'").parse(value)
+                sdf = SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss'Z'")
+                sdf.setTimeZone(TimeZone.getTimeZone("UTC"))
+                parsed_value = sdf.parse(value)
             self._value = parsed_value


wpride:
Also we were using hh date formatting instead of HH which is 12 hour instead of 24 :)

wpride:
I think provided that we can always count on the timeEnd from the meta coming in as UTC this fix will work fine

czue:
ugh. timezones.

> I think provided that we can always count on the timeEnd from the meta coming in as UTC this fix will work fine

So that is happening [here](https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/reports/views.py#L1501). @dannyroberts do you expect `metadata.timeEnd` to always be in UTC? Or do we need to convert it? My guess: we don't need to convert it for domains that have proper timezones set, but do need to convert it for domains that don't.

czue:
(also that hh/HH thing is pretty funny)

dannyroberts:
@wpride I think I need more info about where timeEnd is coming from. If you're parsing it directly out of the XML, then do don't think you have to worry about/consider the timezone bug @czue referenced, as that's a bug with how parses and saves dates from xml on older domains. If you're always working directly from XML then you're fine. If you're working with case.date_opened, and getting that from HQ then it'll be affected.

Can you give me more of an idea of where/how this function is used?

wpride:
@dannyroberts unfortunately don't know that I'm the best to answer that, best I can tell this instance is coming from https://github.com/dimagi/commcare-hq/blob/master/corehq/form_processor/backends/sql/dbaccessors.py#L37

But I'm pretty sure its not coming from the XML directly - from what I can tell we're getting this date (data.meta.timeEnd) from a JSON response

czue:
@dannyroberts It's coming from the form json

snopoke:
@kaapstorm FYI re KERMI timezone issues

wpride:
Removing the "Don't Merge" label but I'm assuming we're waiting on someone (@dannyroberts?)  from HQ team to confirm that these dates will always be UTC. Edit Forms is broken until this is merged tho.

dannyroberts:
OK. I think it would take more of me wrapping my head around this to know for sure the best solution but here are my thoughts
- if you assume that dates in the form JSON are accurate in their use of 'Z' to mean UTC, then it'll definitely be right for all new domains (created in the last ~9 months).
- for older domains, I don't think it's right for touchforms to have to know about HQ's past timezone errors and correct for them; if we're serious about fixing it, we should fix it by having HQ change the dates to the correct UTC ones (to the best of its ability) before sending them to touchforms
- regardless this PR is an obvious improvement.

So with the caveat that if we're serious about fixing the issue for older domains, we'll have to look more into the second bullet, I'm going to merge this. I still don't fully understand what the symptom of the issue is, so I don't know whether it's worth it to track down. Would someone be willing to have a sync on this at some point? Could be in a number of weeks, just feel like it'd be worth either fixing or intentionally not fixing.

czue:
@dannyroberts not sure if you saw the link from my first message but [here it is again](https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/reports/views.py#L1501). Totally agree with you and think we just need to timezone adjust the dates in that code if necessary.

dannyroberts:
Ah thanks for re-posting it. Yes, it looks like those would be pretty easy to adjust. I could do that without more context, but I feel a little funny a couple lines without understanding the context. Is the way editing forms works that it dumps you in CloudCare, and that's why it needs to pass data to touchforms? (Sorry, really noob question, I've just never actually used or thought about our edit-forms function.)

czue:
yeah, so it tells touchforms "go edit this submission, but whenever you need the `now()` function use this date instead" which should be the timeEnd of the form (so that all the logic that depends on `now` doesn't get all screwed up)

dannyroberts:
Cool, thanks!

wpride:
@dannyroberts thank you for the summary and decisive action!

PR #207, opened 2016-03-15
Fix path used for clearing user dbs

wpride:
Fix for http://manage.dimagi.com/default.asp?190831

Really stupid of me, the SQLite DBs were never getting cleared out. This means that closed cases were continuing to live in the DBs. Even though their status was "closed", this would still mess with some intra-form calculations like Jen's (IE if you count all cases of a certain type).


benrudolph:
nice find :mag:

PR #213, opened 2016-04-12
fix firefox again

benrudolph:
@proteusvacuum no idea how this disappeared but this should fix FF issues. possibly a bad merge at one point: https://github.com/dimagi/touchforms/pull/177/files


proteusvacuum:
:fire: :tiger:

PR #214, opened 2016-04-12
fix checkbox spacing

benrudolph:
i've wanted to fix this for so long
before:
screen shot 2016-04-12 at 4 18 45 pm
after:
screen shot 2016-04-12 at 4 33 02 pm


benrudolph:
cc: @sravfeyn

PR #215, opened 2016-04-14
Speed up form submissions for SQL backend

benrudolph:
@wpride did some looking into why form submissions were slow. I'm pretty sure this is the cause. i just straight up removed it because i couldn't figure out what it was doing and it worked fine when i removed it. let me know if you think that is wrong

cc: @sravfeyn

### DIFF
--- a/touchforms/backend/xformplayer.py
+++ b/touchforms/backend/xformplayer.py
@@ -725,14 +725,6 @@ def submit_form(xform_session, answers, prevalidated):
     else:
         resp = form_completion(xform_session)
         resp['status'] = 'success'
-        xml = xform_session.output()
-        if xform_session.uses_sql_backend:
-            process_form_xml(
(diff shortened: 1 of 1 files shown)


wpride:
@benrudolph this is the code that actually causes the new form to be processed in the user's CaseDB. IE if they close a case in the form, this will close that case in the SQLiteDB. Without this you'd need to sync every time you made a transaction.

czue:
any way we could move this asynchronously?

benrudolph:
hmm, that's unfortunate. yeah this would be a perfect case for something like celery. we don't have that infrastructure in the old touchforms, but will mentioned possibly using threading which might work

benrudolph:
@wpride for now i just updated it with a comment explaining what that portion is used for. agree that is should be async though. until we figure out why that's slow i'm imagining we won't be able to resolve slow submits

PR #216, opened 2016-04-28
Ko 3.3

benrudolph:
@orangejenny this should fix touchforms for KO 3.3. there was a super annoying bug in 2.x that didn't allow integers for select values. however when they supported it in 3.3, my previous solution broke haha

cc: @millerdev 

http://manage.dimagi.com/default.asp?225344


touchforms/formplayer/templates/formplayer/fullform-ui/templates.html
```diff
@@ -308,13 +308,12 @@ <h3 class="caption" data-bind="html: header"></h3>
...
                 <input data-bind="
                     checked: $parent.rawAnswer,
-                    value: $index() + 1,
+                    checkedValue: $index() + 1,
                     attr: {
                         id: 'group-' + $parent.entryId + '-choice-' + $index(),
                         type: $parent.isMulti ? 'checkbox' : 'radio',
                         name: $parent.entryId,
                         class: 'group-' + $parent.entryId,
-                        checked: $parent.isChecked($index() + 1)
```
orangejenny: Looks like `isChecked` can be removed altogether from entrycontrols_full.js

benrudolph: good catch 👍 . updated

PR #221, opened 2016-05-17
  Load forms

wpride:
Let us directly load up the XForm tree (don't need to fetch again)

cross https://github.com/dimagi/commcare-hq/pull/11644


touchforms/formplayer/static/formplayer/script/webformsession.js
```diff
@@ -380,3 +381,9 @@ WebFormSession.prototype.initForm = function(args, $form) {
         self.onload(self, resp);
     });
 };
+
+WebFormSession.prototype.loadDirect = function (resp, $form, initLang) {
```
benrudolph: can you document on what this does or change the name?

PR #222, opened 2016-05-24
Merge utils with Formplayer

emord:
http://manage.dimagi.com/default.asp?225851#1151675

Generally agree with the separation here, but merging these two files as it should solve the issue of async loading from the CDN and they were always loaded together anyways

@benrudolph cc @esoergel


benrudolph:
sounds good

PR #223, opened 2016-05-25
debugger z-index

benrudolph:
@czue this kills two 🐦 with one 🗿 the z-index is enough to float over the black bar and the input groups. removed the border cause it looked weird when over the black bar
http://manage.dimagi.com/default.asp?227878
http://manage.dimagi.com/default.asp?227779

cc: @millerdev


millerdev:
Assuming no tests for this.

millerdev:
canceled travis

czue:
nice thanks

PR #230, opened 2016-06-16
Ensure session id gets set

benrudolph:
@czue here's a test for ensuring that the session id gets set. sorry about that, must've missed that somehow. essentially when you call with `.prototype` the `this` object is incorrect. this bug definitely manifests in a weird way. test fails without your fix

cc: @biyeun


czue:
awesome, thanks for the follow up!

PR #231, opened 2016-06-21
add routing directly from browser to formplayer with auth

wpride:
Pass use_formplayer toggle into webformsession.js, call Formplayer directly with auth cookies 

cross: https://github.com/dimagi/formplayer/pull/49
cross: https://github.com/dimagi/commcare-hq/pull/12186

### DIFF
--- a/touchforms/formplayer/static/formplayer/script/webformsession.js
+++ b/touchforms/formplayer/static/formplayer/script/webformsession.js
@@ -59,6 +59,7 @@ function WebFormSession(params) {
     self.taskQueue = new TaskQueue();
     self.formContext = params.formContext;
     self.domain = params.domain;
+    self.formplayerEnabled = params.formplayerEnabled;
 
     if (params.form_uid) {
         self.formSpec = {type: 'form-name', val: params.form_uid};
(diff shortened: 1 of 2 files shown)


touchforms/formplayer/static/formplayer/script/webformsession.js
```diff
@@ -139,14 +140,33 @@ WebFormSession.prototype.serverRequest = function (requestParams, callback, bloc
...
-            data: JSON.stringify(requestParams),
-            dataType: "text"  // we don't use JSON because of a weird bug: http://manage.dimagi.com/default.asp?190983
-        })
-        .success(function(resp) { self.handleSuccess(JSON.parse(resp), callback); })
-        .fail(function (resp, textStatus) { self.handleFailure(JSON.parse(resp), textStatus); });
+
+    if(self.formplayerEnabled){
+        $.ajax({
+                type: 'POST',
+                url: 'http://localhost:8090/' + requestParams.action,
```
snopoke: pretty sure this shouldn't be `localhost`

snopoke: Are we routing this through nginx?

benrudolph: yes, this should be the the FORMPLAYER_URL

benrudolph: also a fair bit of this can be moved after the else clause the success and fail functions are the same and many of the options are as well

benrudolph:
👍 still think the JS can be condensed a bit

PR #233, opened 2016-06-27
use the username and domain for restore

benrudolph:
@snopoke @wpride use username and domain from the request to do the restore


touchforms/formplayer/api.py
```diff
@@ -318,7 +318,7 @@ def sync_db(username, domain=None, auth=None):
     data = {
         "action":"sync-db",
         "username": username,
-        "domain:": domain
+        "domain": domain
```
benrudolph: this was a nice little gotcha 😖

benrudolph:
@wpride i'm still seeing these errors in the logs:
```
org.commcare.modern.models.RecordTooLargeException: You tried to restore some data sized 1.0 MB which exceedsthe maximum allowed size of 1MB. Please reduce the size of this fixture.
	at org.commcare.modern.database.DatabaseHelper.addDataToValues(DatabaseHelper.java:99)
	at org.commcare.modern.database.DatabaseHelper.getMetaFieldsAndValues(DatabaseHelper.java:91)
```
is that expected?

touchforms/backend/xformserver.py
```diff
@@ -259,10 +259,9 @@ def handle_request(content, server):
             result = xfsess.evaluate_xpath(content['xpath'])
             return {"output": result['output'], "status": result['status']}
         elif action == xformplayer.Actions.SYNC_USER_DB:
-            ensure_required_params(['username', 'hq_auth'], action, content)
-            trimmed_username = content['username'][:content['username'].index('.')]
-            result = touchcare.force_ota_restore(trimmed_username, auth=content['hq_auth'])
-            return result
+            ensure_required_params(['username', 'hq_auth', 'domain'], action, content)
+            trimmed_username = content['username'].split('@')[0]
```
wpride: If its a web user this will be wrong, right? IE my username is still wpride@dimagi.com for the purposes of authing

wpride:
@benrudolph Yes, that's expected - SQLite can't store a blob greater than 1MB. I have a backlog item to switch to storing this in the file system for CloudCare (which @phillipm has implemented in Android)

touchforms/backend/xformserver.py
benrudolph: au contraire mr pride, this does in fact work.  i didn't know this until you mentioned it and i looked into why it was succeeding for me. i think when we supply a user with `brudolph@wrong-domain` it uses that wrong domain and then verifies incorrectly. the view code gets the user from the hq_auth so the user isn't incorrect. so that's why it seems that `brudolph@correct-domain` works

benrudolph: (probably still should use yours though)

benrudolph:
closing in favor of: https://github.com/dimagi/touchforms/pull/234

PR #234, opened 2016-06-27
Wp/touchforms sqlite web user

wpride:
We needed to add the domain to the sync-db request so that web users could also sync. Took the opportunity to clean this up a bit for both types.

See http://manage.dimagi.com/default.asp?230606
Branch from https://github.com/dimagi/touchforms/compare/sk/touchforms-sqlite-web-user?expand=1

### DIFF
--- a/touchforms/formplayer/api.py
+++ b/touchforms/formplayer/api.py
@@ -318,7 +318,7 @@ def sync_db(username, domain=None, auth=None):
     data = {
         "action":"sync-db",
         "username": username,
-        "domain:": domain
+        "domain": domain
     }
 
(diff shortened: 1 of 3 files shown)


touchforms/formplayer/api.py
```diff
@@ -318,7 +318,7 @@ def sync_db(username, domain=None, auth=None):
     data = {
```
benrudolph: any reason why we can't make the function signature here `sync_db(username, domain, auth)`? it seems like it always gets called with those and it shouldn't be optional

snopoke:
:+1:

PR #235, opened 2016-06-29
Don't force sync on each filter cases call

wpride:
@benrudolph @snopoke think I only added this param in here for testing purposes, should probably just be refactored out (can't think of a situation when you'd want to enforce this on a get_cases call)


benrudolph:
if it's not being used anywhere, i'd be in favor of dropping it YAGNI!

PR #236, opened 2016-07-13
Fix up dates on cloudcare

benrudolph:
@wpride @czue this should fix the timezone issues india has been seeing. i'm going to explain why i think that just so i've convinced myself. here's the original ticket: http://manage.dimagi.com/default.asp?231447. 

Problem:
In india (or anywhere not in UTC) would experience an issue where constraints on dates would not act as expected. For example the requirement would be for the date to not be past "today" and you couldn't set the date to "today"

This can happen when you're in India and it's July 13th early morning and in UTC it's July 12th late at night. So when you fill out a date question, the current code just strips the timezone and takes the date. This results in a date of 2016-07-13. This'll validate incorrectly on the server side because it uses UTC. If we convert the client side date to UTC when setting the answer, it will then become 2016-07-12 which _will_ validate correctly.

...hate timezones


touchforms/formplayer/static/formplayer/script/entrycontrols_full.js
```diff
@@ -290,32 +290,40 @@ function DateEntry(question, options) {
...
+    var date = moment.utc(serverDate, DateEntry.serverFormat);
+    return date.format(DateEntry.clientFormat);
 };
 
+DateEntry.convertToUTCAnswer = function(rawDate) {
+    if (!rawDate) {
+        return null;
+    }
+    return moment.utc(rawDate).format(DateEntry.serverFormat);
+}
```
dimagimon: Missing semicolon.

touchforms/spec/javascripts/entrycontrols_full_spec.js
```diff
@@ -137,6 +141,15 @@ describe('Entries', function() {
         expect(clientDate).toBe('10/08/2015');
     });
 
+    it('Should correctly handle different timezones', function() {
+        // Late at night in LA (10 pm)
+        var LAdate = moment.tz("2016-12-01T22:00:00", "America/Los_Angeles")
```
dimagimon: Missing semicolon.

touchforms/spec/javascripts/entrycontrols_full_spec.js
```diff
@@ -137,6 +141,15 @@ describe('Entries', function() {
...
+        expect(LAdate.format(DateEntry.serverFormat)).toBe('2016-12-01');
+
+        // Since it's 10pm in LA that means on conversion to UTC, it should be the next day
+        expect(DateEntry.convertToUTCAnswer(LAdate.toDate())).toBe('2016-12-02');
+    })
```
dimagimon: Missing semicolon.

benrudolph:
cc: @orangejenny

czue:
seems logical to me. though i'm curious about one thing - where do the time and timezone in the datepicker come from? does it just use the browser's notion of current settings?

benrudolph:
that's correct, from the browser's timezone settings

czue:
and the current time, i guess?

benrudolph:
yep

PR #241, opened 2016-07-20
Revert "Fix up dates on cloudcare"

benrudolph:
Reverts dimagi/touchforms#236


benrudolph:
this was pretty badly wrong

benrudolph:
going to deploy

benrudolph:
Ok, the reason this guy was bad was I didn't realize that when you pulled the date from the date picker it didn't include the time with it (not sure how I missed that). Anyways that meant the date always was at midnight. When it'd do the UTC conversion on the date, it would then always change the date to the previous day (depending on where you were in relation to UTC). No one has complained about it, but I imagine there might be a few domains that were affected.

The proper solution for this is as follows:

1. Record the date in the local timezone (and keep that information for the backend to use)
2. On the server side, validate the condition either by converting that date to UTC or telling the phone logic what the timezone is.
3. Storing the date in its original timezone

My solution before, even if the datepicker included date and time didn't make that much sense because it essentially stored all dates in UTC and that doesn't make a whole lot of sense when you're looking at form data. There's really no perfect way to solve this without some confusion, but I think it makes most sense to store the date in the timezone that it was recorded in because that's what a phone would do

@czue / @wpride

PR #242, opened 2016-07-20
Update localsettings.py.example

NoahCarnahan:
The value for `POSTGRES_JDBC_JAR` in settings.py will be fine for a normal installation, and the value of this line in localsettings.py.example makes it look like "{COMMCARE_HOME}" will be replaced programatically, but it won't!

@wspride @benrudolph cc @emord


benrudolph:
rough 👍

PR #243, opened 2016-07-22
only validate case xml for sql users

wpride:
@benrudolph offlined, going to make old change in separate PR


benrudolph:
merging since this is a revert and we need it in

PR #245, opened 2016-08-04
Allow minus and periods in numeric question

benrudolph:
@czue @wpride 

http://manage.dimagi.com/default.asp?233937

cc: @orangejenny

### DIFF
--- a/touchforms/formplayer/static/formplayer/script/entrycontrols_full.js
+++ b/touchforms/formplayer/static/formplayer/script/entrycontrols_full.js
@@ -201,7 +201,7 @@ function PhoneEntry(question, options) {
 
     this.getErrorMessage = function(rawAnswer) {
         if (rawAnswer === '') { return null; }
-        return (!(/^\+?[0-9]+$/.test(rawAnswer)) ? "This does not appear to be a valid phone/numeric number" : null);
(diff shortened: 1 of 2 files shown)


touchforms/formplayer/static/formplayer/script/entrycontrols_full.js
```diff
@@ -201,7 +201,7 @@ function PhoneEntry(question, options) {
 
     this.getErrorMessage = function(rawAnswer) {
         if (rawAnswer === '') { return null; }
-        return (!(/^\+?[0-9]+$/.test(rawAnswer)) ? "This does not appear to be a valid phone/numeric number" : null);
+        return (!(/^\+?-?[0-9.]+$/.test(rawAnswer)) ? "This does not appear to be a valid phone/numeric number" : null);
```
orangejenny: It's a little weird that this allows values like `-....3` and `+-3` and `3.`

How about something like `/^[+\-]?\d*(\.\d+)?$/`?

czue:
agree with @orangejenny but :+1: from me otherwise

touchforms/formplayer/static/formplayer/script/entrycontrols_full.js
benrudolph: 👍 will update

PR #246, opened 2016-08-05
CommCare 2.29 jars

phillipm:
The `commcare-api.jar` contains javarosa, sqlite-jdbc, and kxml sources in them, so I'm removing those jars. Not sure if that will break anything...

PR #248, opened 2016-08-08
don't call toString if arg isn't a date

wpride:
Possible fix here: http://manage.dimagi.com/default.asp?234276

Not sure what's going on here exactly - in this form all three calls to this function have a date as their first argument, so not clear why we're receiving a string here. Could be some weird casting going on, could be that the date() method is erroring and returning a blank string.

Either way this will be fixed for real when we merge 2.29 jars here https://github.com/dimagi/touchforms/pull/246


benrudolph:
thanks for the fix

PR #249, opened 2016-08-12
Better error handling

benrudolph:
@wpride formplayer returns a proper json error (unlike old touchforms) so we don't need to do json.parse

cc: @czue


touchforms/formplayer/static/formplayer/script/webformsession.js
```diff
@@ -216,7 +216,7 @@ WebFormSession.prototype.handleFailure = function(resp, textStatus) {
     this.onerror({
         human_readable_message: errorMessage
     });
-    this.onLoadingComplete(true);
+    this.onLoadingComplete();
```
benrudolph: the `true` argument fires off another onerror event so i removed that

czue:
seems good to me

wpride:
✅ from this guy

benrudolph:
let's go for it!

PR #251, opened 2016-08-24
jQuery 3 compatibility

orangejenny:
Backwards-compatible, should be good to merge.

@benrudolph 
cc @kaapstorm / @NoahCarnahan

### DIFF
--- a/touchforms/formplayer/static/formplayer/script/deprecate/entrycontrols_type.js
+++ b/touchforms/formplayer/static/formplayer/script/deprecate/entrycontrols_type.js
@@ -355,9 +355,9 @@ function MultiSelectEntry (args) {
     return function () {
       var cbox = $('#ch-' + i);
       if (cbox.is(':checked')) {
-        cbox.removeAttr('checked');
+        cbox.prop('checked', false);
       } else {
-        cbox.attr('checked', true);
+        cbox.prop('checked', true);
       }
       cbox.focus();
     }
--- a/touchforms/formplayer/static/formplayer/script/webformsession.js
+++ b/touchforms/formplayer/static/formplayer/script/webformsession.js
@@ -142,31 +142,35 @@ WebFormSession.prototype.serverRequest = function (requestParams, callback, bloc
     this.numPendingRequests++;
     this.onLoading();
 
-    if(self.formplayerEnabled){
+    if (self.formplayerEnabled){
         $.ajax({
-                type: 'POST',
-                url: url + "/" + requestParams.action,
-                data: JSON.stringify(requestParams),
-                contentType: "application/json",
-                dataType: "json",
-                crossDomain: {crossDomain: true},
-                xhrFields: {withCredentials: true},
-            })
-            .success(function (resp) {
+            type: 'POST',
+            url: url + "/" + requestParams.action,
+            data: JSON.stringify(requestParams),
+            contentType: "application/json",
+            dataType: "json",
+            crossDomain: {crossDomain: true},
+            xhrFields: {withCredentials: true},
+            success: function(resp) {
                 self.handleSuccess(resp, callback);
-            })
-            .fail(function (resp, textStatus) {
+            },
+            error: function(resp, textStatus) {
                 self.handleFailure(resp, textStatus);
-            });
-    } else{
+            },
+        });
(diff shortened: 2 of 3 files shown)


NoahCarnahan:
👍 on ✅

touchforms/formplayer/static/formplayer/script/deprecate/entrycontrols_type.js
```diff
@@ -355,9 +355,9 @@ function MultiSelectEntry (args) {
     return function () {
       var cbox = $('#ch-' + i);
       if (cbox.is(':checked')) {
-        cbox.removeAttr('checked');
+        cbox.prop('checked', false);
```
benrudolph: i don't think entrycontrols_type.js or typeforms-ui.js is used anymore, but good find! i was scared to delete it so i moved it to the deprecate folder

touchforms/formplayer/static/formplayer/script/webformsession.js
```diff
@@ -142,31 +142,35 @@ WebFormSession.prototype.serverRequest = function (requestParams, callback, bloc
...
-                data: JSON.stringify(requestParams),
-                dataType: "text"  // we don't use JSON because of a weird bug: http://manage.dimagi.com/default.asp?190983
-            })
-            .success(function(resp) { self.handleSuccess(JSON.parse(resp), callback); })
-            .fail(function (resp, textStatus) { self.handleFailure(JSON.parse(resp), textStatus); });
+            type: 'POST',
+            url: url,
+            data: JSON.stringify(requestParams),
+            dataType: "text",  // we don't use JSON because of a weird bug: http://manage.dimagi.com/default.asp?190983
+            success: function(resp) {
```
benrudolph: does jquery no longer support promises?

orangejenny: It's just the success/error/complete functions being removed (https://jquery.com/upgrade-guide/3.0/#breaking-change-special-case-deferred-methods-removed-from-jquery-ajax). I could have left the `fail` call here alone, but my understanding is that moving that code into an error callback instead will behave the same way and is the more common way that HQ code is written.

3.0 does make a bunch of changes to Deferred objects (https://jquery.com/upgrade-guide/3.0/#deferred). They're some of the changes I'm most worried about in upgrading...we don't do a lot of promise-based code, but it's difficult to look at our code and know if it'll be affected.

touchforms/formplayer/static/formplayer/script/deprecate/entrycontrols_type.js
orangejenny: Oh, I didn't even notice it was in a deprecate folder. Been doing a lot of grepping.

I'm a little confused about moving things to a deprecated folder. If it turns out that this is still used somewhere, won't moving it break things anyway?

touchforms/formplayer/static/formplayer/script/webformsession.js
benrudolph: ah i see so `success` is being removed so we either need to switch it to `done` or do what you've done. ouf yeah i didn't read through all the changes they made to promises but that does seem ominous. from a glance it seems that the risk is in handling errors, which is something we don't really do well anyways

touchforms/formplayer/static/formplayer/script/deprecate/entrycontrols_type.js
benrudolph: yeah i also updated the reference to point to deprecate as well. it was more so i could figure out which files were relevant since there were ~9 files that had their code almost duplicated in another file named very similarly. was very confusing. no need to worry about it, i'll take care of deleting them. we'll be removing this submodule soon anyways

orangejenny: Aha, got it, glad you're on top of it :)

touchforms/formplayer/static/formplayer/script/webformsession.js
orangejenny: > from a glance it seems that the risk is in handling errors, which is something we don't really do well anyways

yep, that's exactly it, for better or worse :-/

PR #253, opened 2016-08-30
Restyle to use bootstrap collapsable in debugger

benrudolph:
@wpride @biyeun this makes the debugger somwhat compatible in preview mode. still slightly janky:

screen shot 2016-08-30 at 12 05 34 pm
screen shot 2016-08-30 at 12 05 25 pm
screen shot 2016-08-30 at 12 02 19 pm


benrudolph:
cc: @dannyroberts

PR #256, opened 2016-09-01
Remove logic around presentation group

benrudolph:
@wpride @czue we have this weird branching logic around whether or not a group has a label or not. i can't tell why it was ever used, maybe you guys have more insight? it fixes the issue where TF crashes if repeats don't have labels (much more prevalent now that that's the default)

cc: @dannyroberts

### DIFF
--- a/touchforms/backend/xformplayer.py
+++ b/touchforms/backend/xformplayer.py
@@ -322,11 +322,9 @@ class XFormSession(object):
             evt = self.__parse_event(form_ix)
             evt['relevant'] = relevant
             if evt['type'] == 'sub-group':
-                presentation_group = (evt['caption'] != None)
-                if presentation_group:
-                    siblings.append(evt)
-                    evt['children'] = []
-                form_ix = self._walk(form_ix, evt['children'] if presentation_group else siblings)
+                siblings.append(evt)
+                evt['children'] = []
+                form_ix = self._walk(form_ix, evt['children'])
             elif evt['type'] == 'repeat-juncture':
                 siblings.append(evt)
                 evt['children'] = []


benrudolph:
http://manage.dimagi.com/default.asp?236034

czue:
I have no insight. this is one of the dark and scary places of touchforms I never ventured. probably only drew that knows.

touchforms/backend/xformplayer.py
```diff
@@ -322,11 +322,9 @@ def ix_in_scope(form_ix):
...
             evt['relevant'] = relevant
             if evt['type'] == 'sub-group':
-                presentation_group = (evt['caption'] != None)
-                if presentation_group:
-                    siblings.append(evt)
-                    evt['children'] = []
-                form_ix = self._walk(form_ix, evt['children'] if presentation_group else siblings)
+                siblings.append(evt)
+                evt['children'] = []
+                form_ix = self._walk(form_ix, evt['children'])
```
czue: I stared at this for a while and can't figure out what cases are different and how. do you have a summary? like why you want want to call `self._walk` with `siblings` and whether it's a problem that that no longer happens...

czue:
didn't merge just because i don't understand the change very well, but happy for you to merge if you're confident.

wpride:
@benrudolph Made the corresponding change for formplayer here and confirmed it works in a few cases: https://github.com/dimagi/commcare-core/commit/86c07b082a82bc2470d60596ef527c2a7965562f

touchforms/backend/xformplayer.py
benrudolph: so my impression is that when you call it with "siblings" it changes the structure of the return object to to be a flat list:
 
with presentation group:
```
question1
repeat
  question2
  question3
```

without a presentation group it changes it to a flat list:
```
question1
question2
question3
```
but this breaks TF because the indexes are now off. i have no idea if this was ever working. it's possible that i broke it a long time ago when i refactored TF frontend and it was no longer able to handle that case. groups without labels were probably just extremely rare. either way it seems silly to keep around

benrudolph:
i feel good about this because whenever it executes that code path, it breaks TF so removing it can only be positive (famous last words)

wpride:
once more unto the breach

benrudolph:
when the blast of war blows in our ears,
then imitate the action of the tiger

PR #255, opened 2016-08-31
Revert "Revert "add controls for in-form navigation""

benrudolph:
Reverts dimagi/touchforms#254

@biyeun this is the revert revert

PR #257, opened 2016-09-03
Revert "Revert "Revert "add controls for in-form navigation"""

biyeun:
Reverts dimagi/touchforms#255

@benrudolph ugh i misread the title of the last pr. lol


benrudolph:
haha ok i think we're good now. we should deploy this soon as we're getting errors on prod

PR #263, opened 2016-09-13
assign on_delete explicitly

nickpell:
Get rid of warnings like

```RemovedInDjango20Warning: on_delete will be a required arg for ForeignKey in Django 2.0. Set it to models.CASCADE on models and in existing migrations if you want to maintain the current default behavior. See https://docs.djangoproject.com/en/1.9/ref/models/fields/#django.db.models.ForeignKey.on_delete```

that appear when running django 1.9.  Although it's not needed until django 2.0, I'd just like to see less warnings now.

@sravfeyn @dannyroberts

### DIFF
--- a/touchforms/formplayer/migrations/0001_initial.py
+++ b/touchforms/formplayer/migrations/0001_initial.py
@@ -22,7 +22,7 @@ class Migration(migrations.Migration):
                 ('session_name', models.CharField(max_length=100)),
                 ('created_date', models.DateTimeField(default=datetime.datetime.utcnow)),
                 ('last_activity_date', models.DateTimeField(null=True)),
-                ('user', models.ForeignKey(to=settings.AUTH_USER_MODEL)),
+                ('user', models.ForeignKey(to=settings.AUTH_USER_MODEL, on_delete=models.CASCADE)),
             ],
             options={
             },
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: touchforms/formplayer/migrations/0001_initial.py [hunk 1]
```diff
@@ -22,7 +22,7 @@ class Migration(migrations.Migration):
                 ('session_name', models.CharField(max_length=100)),
                 ('created_date', models.DateTimeField(default=datetime.datetime.utcnow)),
                 ('last_activity_date', models.DateTimeField(null=True)),
-                ('user', models.ForeignKey(to=settings.AUTH_USER_MODEL)),
+                ('user', models.ForeignKey(to=settings.AUTH_USER_MODEL, on_delete=models.CASCADE)),
```
sravfeyn: Not sure about the behaviour of passing `on_delete` kwarg, but if it's just to get rid of warnings it might be worth to add a comment here to say so (so that it can reviewed when upgrading to Django 2.0 latter)?

nickpell: The behavior is unchanged, since ```CASCADE``` is the default value for ```on_delete```: https://docs.djangoproject.com/en/1.7/ref/models/fields/#django.db.models.CASCADE

Later when django gets upgraded to 2.0, explicitly setting ```on_delete``` will be required, so this change can stay as is at that time.

sravfeyn: 👍

PR #265, opened 2016-09-20
update form styles

biyeun:


benrudolph:
this should now be in the commcarehq: https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/cloudcare/templates/form_entry/templates.html

benrudolph:
surprised it was working (was a fairly recent change though). going to delete these ones once i'm confident nothing is requesting them

PR #269, opened 2016-10-27
Add hardcoded array source for nepali and ethiopian months, remove hack

wpride:
Fixes http://manage.dimagi.com/default.asp?239754#edit_1_239754

First remove the old IHandler hack that was just returning the unformatted date
Next, we need to register a new data source for the months arrays since old CloudCare won't have the localizations (which I forgot to account for originally)

@benrudolph

### DIFF
--- a/touchforms/backend/customhandlers.py
+++ b/touchforms/backend/customhandlers.py
@@ -63,7 +63,7 @@ def attach_handlers(form, extensions, context, preload_data=None):
                         form.exprEvalContext.addFunctionHandler(handler)
 
 
-# We need this for old CloudCare, which doesn't have localizations 
+# We need this for old CloudCare, which doesn't have localizations
 class HardCodedArrayDataSource(ArrayDataSource):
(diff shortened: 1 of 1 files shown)


touchforms/backend/customhandlers.py
```diff
@@ -60,31 +62,14 @@ def attach_handlers(form, extensions, context, preload_data=None):
...
-
-    def getName(self):
-        return "format-date-for-calendar"
-
-    def eval(self, args, ec):
-        if isinstance(args[0], java.util.Date):
-            return args[0].toString()
-        return args[0]
-
+class HardCodedArrayDataSource(ArrayDataSource):
```
dimagimon: E302 expected 2 blank lines, found 1

benrudolph:
dang thanks for digging. i think in the future let's just punt/try and convert them to new cloudcare if we know that fixes it

benrudolph:
(though edit for still doesn't work, but should soon)

PR #280, opened 2016-12-07
Support barcode in TF

benrudolph:
@wpride since we know support barcode as a string in formplayer we need to support it in TF so we can edit forms: http://manage.dimagi.com/default.asp?243542

cc: @emord


wpride:
I'm will pride and I approve this pull request

PR #287, opened 2017-03-14
fix bad usage of dest

nickpell:
http://stackoverflow.com/questions/18335687/python-using-argparse-argumentparser-method?answertab=votes#tab-top

@mkangia


benrudolph:
can just remove, never used

dimagimon:
Could not review pull request. It may be too large, or contain no reviewable changes.

nickpell:
thanks @benrudolph updated

PR #290, opened 2017-09-08
ElementTree -> cElementTree

millerdev:
See https://github.com/dimagi/commcare-hq/pull/17756

@dannyroberts @snopoke cc @calellowitz


dannyroberts:
fyi touchforms is barely used anymore; it's still around I think only for playing forms over SMS

PR #291, opened 2017-10-25
add py 3.x to test matrix

nickpell:


.travis.yml
```diff
@@ -1,6 +1,9 @@
 language: python
 python:
   - "2.7"
+  - "3.4"
```
emord: I don't think this is real. https://github.com/dimagi/touchforms/blob/master/run_tests.sh uses jython to actually run the code. jython 3 is in development

nickpell: ah I see

PR #292, opened 2018-01-31
SMS for Formplayer

wpride:
Updates to Formplayer to handle SMS requests

This PR

- Adds infrastructure for an experiment to compare Formplayer and Touchforms responses
- `nav_mode`, `restore_as`, and `sessionid` cookie to Formplayer requests

Depends on:
- https://github.com/dimagi/commcare-cloud/pull/1323
- https://github.com/dimagi/commcare-hq/pull/19319
- https://github.com/dimagi/formplayer/pull/607

### DIFF
--- a/touchforms/formplayer/api.py
+++ b/touchforms/formplayer/api.py
@@ -264,7 +264,7 @@ def post_data_helper(d, auth, content_type, url):
     headers["Set-Cookie"] = 'sessionid=rl5dzcd02npn2qhn6aw732q17t7oc5rf'
     headers["_cookie"] = 'sessionid=rl5dzcd02npn2qhn6aw732q17t7oc5rf'
     headers["Cookie"] = 'sessionid=rl5dzcd02npn2qhn6aw732q17t7oc5rf'
-
+    logging.info('Post AUTH %s' % auth)
     conn = httplib.HTTPConnection(up.netloc)
     conn.request('POST', up.path, data, headers)
     resp = conn.getresponse()
@@ -282,6 +282,7 @@ def post_data(data, auth=None, content_type="application/json"):
     if auth:
         d['hq_auth'] = auth.to_dict()
     domain = d.get("domain")
+    d['restoreAs'] = d['username']
 
     if domain:
         d['uses_sql_backend'] = use_sqlite_backend(domain)
--- a/touchforms/formplayer/api.py
+++ b/touchforms/formplayer/api.py
@@ -301,7 +301,7 @@ def get_candidate_session_data(control_data):
 
     if candidate_session_id is None:
         logging.info("Could not get Formplayer session_id for Touchforms session_id %s" % control_session_id)
-        return (False, None)
+        return False, None
 
     candidate_data["session_id"] = candidate_session_id
     candidate_data["session-id"] = candidate_session_id
--- a/touchforms/formplayer/api.py
+++ b/touchforms/formplayer/api.py
@@ -260,6 +260,7 @@ class XformsResponse(object):
 def post_data_helper(d, auth, content_type, url):
     d['nav_mode'] = 'prompt'
     data = json.dumps(d)
+    logging.info("posting data %s " % data)
     up = urlparse(url)
     headers = {}
     headers["content-type"] = content_type
@@ -280,8 +281,6 @@ def post_data(data, auth=None, content_type="application/json"):
     if auth:
         d['hq_auth'] = auth.to_dict()
     domain = d.get("domain")
-    if domain:
-        d['uses_sql_backend'] = use_sqlite_backend(domain)
     if 'username' in d:
         d['restoreAs'] = d['username']
     # just default to old server for now
--- a/touchforms/formplayer/api.py
+++ b/touchforms/formplayer/api.py
@@ -293,19 +293,20 @@ def perform_experiment(d, auth, content_type):
     with experiment.candidate() as c:
         # If we should already have a session, look up its ID in the experiment mapping. it better be there.
         # This is terrible, but we use both in different places.
-        cache = get_redis_client()
+        control_session_id = None
+
         if "session_id" in d:
             control_session_id = d["session_id"]
-            candidate_session_id = cache.get('touchforms-to-formplayer-session-id-%s' % control_session_id)
-            d["session_id"] = candidate_session_id
-            d["session-id"] = candidate_session_id
-            logging.info("Got candidate %s from control %s" % (candidate_session_id, control_session_id))
         if "session-id" in d:
             control_session_id = d["session-id"]
+
+        if control_session_id is not None:
+            cache = get_redis_client()
             candidate_session_id = cache.get('touchforms-to-formplayer-session-id-%s' % control_session_id)
-            d["session-id"] = candidate_session_id
-            d["session_id"] = candidate_session_id
-            logging.info("Got candidate %s from control %s" % (candidate_session_id, control_session_id))
+            if candidate_session_id is not None:
+                d["session_id"] = candidate_session_id
+                d["session-id"] = candidate_session_id
+
         d['oneQuestionPerScreen'] = True
         c.record(post_data_helper(d, auth, content_type, settings.FORMPLAYER_URL + "/" + d["action"]))
     objects = experiment.run()
--- a/touchforms/formplayer/api.py
+++ b/touchforms/formplayer/api.py
@@ -7,6 +7,7 @@ import socket
 
 from corehq.form_processor.utils.general import use_sqlite_backend
 from touchforms.formplayer.exceptions import BadDataError
+from experiments import FormplayerExperiment
 """
 A set of wrappers that return the JSON bodies you use to interact with the formplayer
 backend for various sets of tasks.
@@ -287,8 +288,23 @@ def post_data(data, auth=None, content_type="application/json"):
     if domain:
         d['uses_sql_backend'] = use_sqlite_backend(domain)
     # just default to old server for now
-    url = settings.XFORMS_PLAYER_URL + '/' + d['action']
-    return post_data_helper(d, auth, content_type, url)
+    return perform_experiment(d, auth, content_type)
+
+
+def perform_experiment(d, auth, content_type):
+    experiment = FormplayerExperiment(name=d["action"], context={'request': d})
+    with experiment.control() as c:
+        c.record(post_data_helper(d, auth, content_type, settings.XFORMS_PLAYER_URL))
+    with experiment.candidate() as c:
+        # If we should already have a session, look up its ID in the experiment mapping. it better be there.
+        # This is terrible, but we use both in different places.
+        if "session_id" in d:
+            d["session_id"] = FormplayerExperiment.session_id_mapping.get(d["session_id"])
+        if "session-id" in d:
+            d["session-id"] = FormplayerExperiment.session_id_mapping.get(d["session-id"])
+        c.record(post_data_helper(d, auth, content_type, settings.FORMPLAYER_URL + "/" + d["action"]))
+    objects = experiment.run()
+    return objects
 
 
 def get_response(data, auth=None):
--- a/touchforms/formplayer/api.py
+++ b/touchforms/formplayer/api.py
@@ -65,7 +65,7 @@ class XFormsConfig(object):
     
     def __init__(self, form_path=None, form_content=None, language="", 
                  session_data=None, preloader_data={}, instance_content=None,
-                 touchforms_url=None, auth=None):
+                 touchforms_url=None, auth=None, restore_as=None, restore_as_case_id=None):
         
         if bool(form_path) == bool(form_content):
             raise XFormsConfigException\
@@ -80,6 +80,8 @@ class XFormsConfig(object):
         self.instance_content = instance_content
         self.touchforms_url = touchforms_url or settings.XFORMS_PLAYER_URL
         self.auth = auth
+        self.restore_as = restore_as
+        self.restore_as_case_id = restore_as_case_id
         
     def get_touchforms_dict(self):
         """
@@ -101,6 +103,17 @@ class XFormsConfig(object):
         self.add_key_helper('username', ret)
         self.add_key_helper('domain', ret)
         self.add_key_helper('app_id', ret)
+
+        if self.restore_as_case_id:
+            # The contact starting the survey is a case who will be
+            # filling out the form for itself.
+            ret['restoreAsCaseId'] = self.restore_as_case_id
+        elif self.restore_as:
+            # The contact starting the survey is a user.
+            ret['restoreAs'] = self.restore_as
+        else:
+            raise ValueError("Unable to determine 'restore as' contact for formplayer")
+
         return ret
 
     def add_key_helper(self, key, ret):
@@ -286,8 +299,6 @@ def post_data(data, auth=None, content_type="application/json"):
 
     if auth:
         d['hq_auth'] = auth.to_dict()
-    if 'username' in d:
-        d['restoreAs'] = d['username']
     # just default to old server for now
     return perform_experiment(d, auth, content_type)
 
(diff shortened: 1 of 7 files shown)


touchforms/formplayer/tests/test_experiment.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+class ExperimentTest(SimpleTestCase):
+
+    control = {}
+    candidate = {}
+
+    def testExperimentStringCompare(self):
+        # session_id will never be equal, should be ignored
+        assert formplayer_string_compare("123", "234", current_key="session_id")
```
gcapalbo: for tests its preferable to use `self.assertTrue()`, `self.assertFalse`, `self.assertEqual`, etc.

touchforms/formplayer/tests/test_experiment.py
```diff
@@ -0,0 +1,54 @@
...
+        assert formplayer_string_compare("123", "234", current_key="output")
+
+        assert formplayer_string_compare(0, "false", current_key="repeatable")
+        assert formplayer_string_compare(1, "true", current_key="repeatable")
+        assert not formplayer_string_compare(1, "false", current_key="repeatable")
+        assert not formplayer_string_compare(0, "true", current_key="repeatable")
+
+    def testExperimentDictCompare(self):
+
+        self.reset()
```
gcapalbo: usually we prep the test in `setUp()` and clean up in `tearDown()`

touchforms/formplayer/experiments.py
```diff
@@ -0,0 +1,102 @@
...
+                'candidate_duration': candidate.duration
+            })
+
+
+def compare_list(control, candidate):
+    is_equal = True
+    for first_item, second_item in zip(control, candidate):
+        if not formplayer_compare(first_item, second_item):
+            is_equal = False
+    return is_equal
```
gcapalbo: should this also be checking the list length? since if the length differs between the two lists, zip will exclude any extra elements

touchforms/formplayer/experiments.py
```diff
@@ -0,0 +1,102 @@
...
+        timing_logger.info(
+            "MSG",
+            extra={
+                'action': self.name,
+                'control_duration': control.duration,
+                'candidate_duration': candidate.duration
+            })
+
+
+def compare_list(control, candidate):
```
gcapalbo: would be good to also check that candidate is a list before doing the rest of the checks

touchforms/formplayer/experiments.py
```diff
@@ -0,0 +1,102 @@
...
+    is_equal = True
+    for first_item, second_item in zip(control, candidate):
+        if not formplayer_compare(first_item, second_item):
+            is_equal = False
+    return is_equal
+
+
+def compare_dict(control, candidate):
```
gcapalbo: would be good to also check that candidate is a dict before doing the rest of the checks

touchforms/formplayer/experiments.py
```diff
@@ -0,0 +1,102 @@
...
+def formplayer_string_compare(control, candidate, current_key=None):
+    if current_key == "session_id":
+        # clearly these will be different
+        return True
+    elif current_key == "output":
+        # don't have any way to compare the XML output at the moment, esp. considering uuids and times
+        return True
+    elif current_key == "repeatable":
+        # These end up as '0' and '1' in python world, despite being set to True and False in xformplayer.py
+        if control == 0:
```
gcapalbo: should this be int `0` or string `'0'`? wasn't sure from the comment above

touchforms/formplayer/experiments.py
```diff
@@ -0,0 +1,102 @@
...
+
+class FormplayerExperiment(laboratory.Experiment):
+    session_id_mapping = {}
+
+    def publish(self, result):
+        # if we're starting a new form, we need to store the mapping between session_ids so we can use later
+        if (self.name == "new-form"):
+            control_session_id = json.loads(result.control.value)["session_id"]
+            candidate_session_id = json.loads(result.observations[0].value)["session_id"]
+            FormplayerExperiment.session_id_mapping[control_session_id] = candidate_session_id
```
gcapalbo: it looks like FormplayerExperiment is being used by client processes when we post to touchforms, so storing this as a dict on the class won't work because the client can be invoked from different processes which won't share this dict.

i think a better way to store the session id mapping is to use redis. you can set a key in redis to be something like `'touchforms-to-formplayer-session-id-%s' % control_session_id`, set its value to be `candidate_session_id`, and set it to expire in 1 week (the max smsforms session length). then you can just look up `'touchforms-to-formplayer-session-id-%s' % control_session_id` anywhere you need it to get the formplayer session id

touchforms/formplayer/api.py
```diff
@@ -287,8 +288,23 @@ def post_data(data, auth=None, content_type="application/json"):
...
+    with experiment.candidate() as c:
+        # If we should already have a session, look up its ID in the experiment mapping. it better be there.
+        # This is terrible, but we use both in different places.
+        if "session_id" in d:
+            d["session_id"] = FormplayerExperiment.session_id_mapping.get(d["session_id"])
+        if "session-id" in d:
+            d["session-id"] = FormplayerExperiment.session_id_mapping.get(d["session-id"])
+        c.record(post_data_helper(d, auth, content_type, settings.FORMPLAYER_URL + "/" + d["action"]))
+    objects = experiment.run()
+    return objects
```
gcapalbo: is this always returning the value of the control right now?

touchforms/formplayer/api.py
```diff
@@ -287,8 +288,23 @@ def post_data(data, auth=None, content_type="application/json"):
...
+        c.record(post_data_helper(d, auth, content_type, settings.XFORMS_PLAYER_URL))
+    with experiment.candidate() as c:
+        # If we should already have a session, look up its ID in the experiment mapping. it better be there.
+        # This is terrible, but we use both in different places.
+        if "session_id" in d:
+            d["session_id"] = FormplayerExperiment.session_id_mapping.get(d["session_id"])
+        if "session-id" in d:
+            d["session-id"] = FormplayerExperiment.session_id_mapping.get(d["session-id"])
+        c.record(post_data_helper(d, auth, content_type, settings.FORMPLAYER_URL + "/" + d["action"]))
+    objects = experiment.run()
```
gcapalbo: when this gets run, are there side effects from running it in both places? for example, do two forms get submitted, one from each place, if someone gets all the way through a survey?

touchforms/formplayer/experiments.py
gcapalbo: nevermind, saw you added this later

touchforms/formplayer/experiments.py
gcapalbo: nevermind, saw you added this later

touchforms/formplayer/experiments.py
```diff
@@ -59,42 +59,43 @@ def compare_list(control, candidate):
     return is_equal
 
 
-def compare_dict(control, candidate):
+def compare_dict(control, candidate, current_key):
     is_equal = True
     if control is not None and candidate is None:
+        logging.info('Key %s has null candidate but real control %s' % (current_key, control))
```
gcapalbo: i don't think you'll ever hit this because you're checking if candidate is a dict before you call this function

gcapalbo:
@wpride confirming you ran the smsforms test suite with all of the new stuff and it passed?

gcapalbo:
@wpride also i don't remember if i mentioned this but smsforms doesn't require users to make app builds, and it always uses the latest saved version of an app from the app/form builder. is formplayer going to behave the same way for these situations?

wpride:
@gcapalbo No you didn't, I can look into it but think App Preview does the same. Is there any reason for this exception?

gcapalbo:
On Tue, Feb 13, 2018 at 8:35 AM, wpride <notifications@github.com> wrote:

> @gcapalbo <https://github.com/gcapalbo> No you didn't, I can look into it
> but think App Preview does the same. Is there any reason for this exception?
>
Mainly for ease of use. Clients don't expect to have to build an app in
order to use it for SMS surveys

touchforms/formplayer/experiments.py
wpride: good call, thanks

touchforms/formplayer/experiments.py
wpride: `0`, fixing comment

touchforms/formplayer/api.py
wpride: Yes

touchforms/formplayer/api.py
```diff
@@ -283,10 +284,8 @@ def post_data(data, auth=None, content_type="application/json"):
     if auth:
         d['hq_auth'] = auth.to_dict()
     domain = d.get("domain")
-    d['restoreAs'] = d['username']
-
-    if domain:
-        d['uses_sql_backend'] = use_sqlite_backend(domain)
```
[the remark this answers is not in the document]
wpride: Hm, good point - I'd convinced myself that this wasn't needed anymore, but looking again I'm not sure. Reverting

touchforms/formplayer/api.py
wpride: Good question. In the `prompt` mode Formplayer will never actually submit a form as currently configured - the output XML is returned to the calling service when the form reaches the end and will be ignored by the experiment. I assume SMS doesn't support case claim? That's the only other side effect I could think of (besides the creation of Formplayer sessions and user DBs which shouldn't impact anything).

touchforms/formplayer/api.py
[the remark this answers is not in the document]
wpride: @gcapalbo OK I remembered my logic on this. I'm pretty sure this would never be set to `True` for SMS requests because they weren't including the `domain` in the request payload until I [added that](https://github.com/dimagi/touchforms/pull/292/files/8107d2fbf2602075c487e13a7213b1735a0af703#diff-a8233683af9d9f2533f234c6e299320aR99) in this PR. So this would never have been set since SMS is the only thing still using this API. Including the domain and having this be set to `True` now causes problems. Let me know if you disagree with that logic.

wpride:
@gcapalbo Confirmed that forms without builds work; SMS passes the XForm content directly to Formplayer so no app install is even needed.

touchforms/formplayer/experiments.py
```diff
@@ -0,0 +1,151 @@
...
+            ret = control == candidate
+    else:
+        ret = control == candidate
+    if not ret:
+        logging.info("Mismatch with key %s between control %s and candidate %s"
+                     % (key, control, candidate))
+    return ret
+
+## Keys that SMS doesn't use or that we otherwise can skip
+def check_skip_key(key):
```
wpride: @gcapalbo you might have done so on your first pass but would be great if you could double check the assumptions I made here (covered many of them on Slack)

touchforms/formplayer/experiments.py
```diff
@@ -78,7 +78,7 @@ def compare_dict(control, candidate, current_key):
     is_equal = True
     for key in control:
         if check_skip_key(key):
-            return True
+            break
```
gcapalbo: should this `continue` and not `break`?

gcapalbo: nevermind, you fixed it in the next commit

touchforms/formplayer/experiments.py
```diff
@@ -53,6 +53,8 @@ def log_timing(self, control, candidate):
 
 def compare_list(control, candidate):
     is_equal = True
+    if len(control) != len(candidate):
+        logging.info("Length of control %s is different than candidate %s" % (control, candidate))
```
gcapalbo: this should also return False right?

touchforms/formplayer/experiments.py
gcapalbo: yes these are fine

wpride: 👍

touchforms/formplayer/api.py
```diff
@@ -274,12 +280,36 @@ def post_data(data, auth=None, content_type="application/json"):
...
+            candidate_session_id = cache.get('touchforms-to-formplayer-session-id-%s' % control_session_id)
+            d["session_id"] = candidate_session_id
+            d["session-id"] = candidate_session_id
+            logging.info("Got candidate %s from control %s" % (candidate_session_id, control_session_id))
+        if "session-id" in d:
+            control_session_id = d["session-id"]
+            candidate_session_id = cache.get('touchforms-to-formplayer-session-id-%s' % control_session_id)
+            d["session-id"] = candidate_session_id
+            d["session_id"] = candidate_session_id
+            logging.info("Got candidate %s from control %s" % (candidate_session_id, control_session_id))
```
gcapalbo: there seems to be a bit of duplication in these two blocks which would be good if it was consolidated.

also, we can't guarantee that there will be a value for `candidate_session_id`. when this is released there will be open touchforms sessions with no corresponding formplayer session and when people answer the next question in those surveys it will get here with a `candidate_session_id` that is `None`

touchforms/formplayer/api.py
```diff
@@ -279,43 +280,52 @@ def post_data(data, auth=None, content_type="application/json"):
...
-    experiment = FormplayerExperiment(name=d["action"], context={'request': d})
+def get_candidate_session_data(control_data):
+    candidate_data = copy.deepcopy(control_data)
+    candidate_data['oneQuestionPerScreen'] = True
+    if "session_id" in control_data:
+        control_session_id = control_data["session_id"]
+    elif "session-id" in control_data:
+        control_session_id = control_data["session-id"]
+    else:
+        return True, candidate_data
```
gcapalbo: i'm not sure if we'll ever reach this scenario, but it looks like it would currently still run the experiment by passing formplayer whatever session id touchforms used since we didn't overwrite session_id in the `candidate_data` (which seems wrong). should this return `False` instead?

touchforms/formplayer/experiments.py
```diff
@@ -136,15 +137,18 @@ def formplayer_string_compare(control, candidate, key=None):
...
-def check_skip_key(key):
+## Keys that we can't test the content of, but want to be present in the candidate
+def check_skip_content(key):
     if key == "session_id":
         # clearly these will be different
         return True
     elif key == "output":
         # don't have any way to compare the XML output at the moment, esp. considering uuids and times
         return True
-    elif key == "seq_id":
```
gcapalbo: would be good to explicity return False here (it will currently return None which will still work for  this use case but better to be explicit)

touchforms/formplayer/api.py
wpride: This is the case where the request doesn't have any `session_id` associated with it yet (such as `new-form`). I don't think you could reach this block if the request had a touchforms `session_id`?

gcapalbo: i see, i was misreading this

gcapalbo:
i was testing this locally and regression-wise the integration between touchforms and smsforms looks good.

but i'm seeing errors in the experiment for formplayer - you can run the test suite to see them. the main thing is that when it's a case filling out the form, the "restore as a case" feature is not working properly. it's passing the case's name as the username to restore as and formplayer is raising error `Invalid restore as user ...`

wpride:
(working for me locally after merging)

gcapalbo:
when i tested locally i was testing with that change. it looks like [this](https://github.com/dimagi/touchforms/pull/292/files#diff-a8233683af9d9f2533f234c6e299320aR284) is still being set even when it's a case answering the form, and it's set to the case's name. is that wrong?

touchforms/formplayer/api.py
```diff
@@ -90,10 +93,13 @@ def get_touchforms_dict(self):
...
                 ("preloader-data", self.preloader_data),
                 ("session-data", self.session_data),
-                ("lang", self.language))
+                ("lang", self.language),
+                ("form-url", self.form_path))
         
         # only include anything with a value, or touchforms gets mad
-        return dict(filter(lambda x: x[1], vals))
+        ret = dict(filter(lambda x: x[1], vals))
+        ret.update(self.session_data)
```
gcapalbo: i just noticed that this is going to duplicate every item in session_data at the root level of the dict. is this necessary, or can we be more selective about the items that formplayer needs instead of adding everything?

PR #295, opened 2018-02-27
Check for candidate failure

wpride:
Realized looking at the source code that these values will be set if an exception occurs, check for them and log if set.

### DIFF
--- a/touchforms/formplayer/experiments.py
+++ b/touchforms/formplayer/experiments.py
@@ -24,6 +24,9 @@ class FormplayerExperiment(laboratory.Experiment):
             self.log_diff(control, candidate)
             return
 
+        if candidate.failure:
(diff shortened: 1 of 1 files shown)


touchforms/formplayer/experiments.py
```diff
@@ -24,6 +24,9 @@ def publish(self, result):
             self.log_diff(control, candidate)
             return
 
+        if candidate.failure:
+            logging.info("Candidate %s failed with exception %s" % (candidate, candidate.exception))
```
gcapalbo: this isn't using one of the loggers you defined at the top of the file (and there are some other places in here that aren't either). is that intentional?

wpride: Yes - I don't want to log to the diff file in this case, I just want to see the exception that occurred in the machine logs

PR #296, opened 2018-03-01
Use get_nimbus_url() and log URL

wpride:
ibid


gcapalbo:
@wpride were you able to trace through and confirm this is what webapps uses?

PR #298, opened 2018-03-03
Log entire URL string

wpride:
The logging I was doing before didn't have the `scheme` component (noob). the logs currently are showing a `301 moved` response for the requests to Formplayer, which I can reproduce when I have `http` or nothing as the scheme (instead of `https`). @gcapalbo do you think it could be the case that the wrong scheme is being used? Looks like `https` is the `DEFAULT_PROTOCOL` in `localsettings.py` on `hqcelery1` so that seems right

### DIFF
--- a/touchforms/formplayer/api.py
+++ b/touchforms/formplayer/api.py
@@ -283,7 +283,7 @@ def post_data_helper(d, auth, content_type, url, log=False):
     data = json.dumps(d)
     up = urlparse(url)
     if log:
-        logging.info("Request to url: %s%s" % (up.netloc, up.path))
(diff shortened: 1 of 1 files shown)


gcapalbo:
it should be using https and if it's not then that's a bug in the result from [get_nimbus_api](https://github.com/dimagi/commcare-hq/blob/6591cbf1ae9a8b5ea24b6485a50625b628458aba/corehq/apps/nimbus_api/utils.py#L6)

301s can also sometimes be caused when there's a trailing / that's missing. could that be the case here?

gcapalbo:
i just confirmed on prod that get_nimbus_url() does use https, so the 301 must be caused by something else

PR #299, opened 2018-03-05
Separate function for Formplayer requests

wpride:
`post_data_helper` is doing some manipulation with the `urlparse()` object so add a new function modeled on the [validation endpoint](https://github.com/dimagi/commcare-hq/blob/3e88dbe4554316981e7f9a43fde34aa8870702c9/corehq/apps/nimbus_api/form_validation.py#L43-L43)

### DIFF
--- a/touchforms/formplayer/api.py
+++ b/touchforms/formplayer/api.py
@@ -298,7 +298,6 @@ def formplayer_post_data_helper(d, auth, content_type, url):
     up = urlparse(url)
     logging.info("Request to url: %s" % up.geturl())
     headers = {}
-    headers["content-type"] = content_type
     headers["Content-Type"] = content_type
     headers["content-length"] = len(data)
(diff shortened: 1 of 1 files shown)


touchforms/formplayer/api.py
```diff
@@ -279,24 +280,37 @@ def server_down(cls):
...
+
+
+def formplayer_post_data_helper(d, auth, content_type, url):
+    d['nav_mode'] = 'prompt'
+    data = json.dumps(d)
+    up = urlparse(url)
+    logging.info("Request to url: %s" % up.geturl())
+    headers = {}
+    headers["content-type"] = content_type
+    headers["Content-Type"] = content_type
```
gcapalbo: why is this here twice?

wpride: Copied over from the `validate_forms()` but I now read this is case-insensitive

PR #300, opened 2018-03-12
Add toggle to use Formplayer for SMS handling

wpride:
cross-request: https://github.com/dimagi/commcare-hq/pull/19763

### DIFF
--- a/touchforms/formplayer/api.py
+++ b/touchforms/formplayer/api.py
@@ -11,6 +11,7 @@ from corehq.form_processor.utils.general import use_sqlite_backend
 from touchforms.formplayer.exceptions import BadDataError
 from experiments import FormplayerExperiment
 from corehq.apps.nimbus_api.utils import get_nimbus_url
+from corehq import toggles
 import requests
 """
 A set of wrappers that return the JSON bodies you use to interact with the formplayer
@@ -294,6 +295,7 @@ def post_data_helper(d, auth, content_type, url, log=False):
 
 def formplayer_post_data_helper(d, auth, content_type, url):
     d['nav_mode'] = 'prompt'
+    d['oneQuestionPerScreen'] = True
     data = json.dumps(d)
     up = urlparse(url)
     logging.info("Request to url: %s" % up.geturl())
@@ -320,6 +322,10 @@ def post_data(data, auth=None, content_type="application/json"):
     if auth:
         d['hq_auth'] = auth.to_dict()
     # just default to old server for now
+
+    if toggles.SMS_USE_FORMPLAYER:
+        return formplayer_post_data_helper(d, auth, content_type, url)
+
     return perform_experiment(d, auth, content_type)
 
 
--- a/touchforms/formplayer/api.py
+++ b/touchforms/formplayer/api.py
@@ -324,7 +324,7 @@ def post_data(data, auth=None, content_type="application/json"):
     # just default to old server for now
 
     if toggles.SMS_USE_FORMPLAYER:
-        return formplayer_post_data_helper(d, auth, content_type, url)
+        return formplayer_post_data_helper(d, auth, content_type, get_nimbus_url() + "/" + data["action"])
 
     return perform_experiment(d, auth, content_type)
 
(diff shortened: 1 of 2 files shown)


touchforms/formplayer/api.py
```diff
@@ -320,6 +322,10 @@ def post_data(data, auth=None, content_type="application/json"):
     if auth:
         d['hq_auth'] = auth.to_dict()
     # just default to old server for now
+
+    if toggles.SMS_USE_FORMPLAYER:
+        return formplayer_post_data_helper(d, auth, content_type, get_nimbus_url() + "/" + data["action"])
```
gcapalbo: `perform_experiment` calls `get_candidate_session_data` which sets the values of both `session_id` and `session-id` to the same value. is session id already being set to the right param here or do both need to be set here too?

wpride: Great point... also occurs to me that when we switch this over we'll want to continue doing the lookups between `touchforms` and `formplayer` `session_id`s so that we don't break any ongoing sessions

touchforms/formplayer/api.py
```diff
@@ -293,7 +294,7 @@ def post_data_helper(d, auth, content_type, url, log=False):
 
 
 def formplayer_post_data_helper(d, auth, content_type, url):
-    d['nav_mode'] = 'prompt'
+    d = get_formplayer_session_data(d)
```
gcapalbo: it looks like `perform_experiment` calls `get_candidate_session_data` before calling `formplayer_post_data_helper` so it seems like calling `get_formplayer_session_data` is unnecessary in that flow.

should this call to `get_formplayer_session_data` be moved to be directly under the toggle check?

touchforms/formplayer/api.py
```diff
@@ -320,9 +321,34 @@ def post_data(data, auth=None, content_type="application/json"):
     if auth:
         d['hq_auth'] = auth.to_dict()
     # just default to old server for now
+
+    if toggles.SMS_USE_FORMPLAYER:
```
gcapalbo: should check `.enabled()`

touchforms/formplayer/api.py
```diff
@@ -320,9 +320,35 @@ def post_data(data, auth=None, content_type="application/json"):
     if auth:
         d['hq_auth'] = auth.to_dict()
     # just default to old server for now
+
+    if toggles.SMS_USE_FORMPLAYER.enabled():
```
gcapalbo: need to pass the domain in too.

would be good to do a run through of the test suite and confirm that nothing breaks with these changes

touchforms/formplayer/api.py
```diff
@@ -320,8 +321,9 @@ def post_data(data, auth=None, content_type="application/json"):
     if auth:
         d['hq_auth'] = auth.to_dict()
     # just default to old server for now
-
-    if toggles.SMS_USE_FORMPLAYER.enabled():
+    domain = d.get("domain")
```
gcapalbo: where does domain come from? i don't think it's getting passed on every request. when i add these lines here

    if not domain:
            raise RuntimeError()

and run the tests, it raises the RuntimeError

gcapalbo:
this will also need to be updated after https://github.com/dimagi/touchforms/pull/301 is merged

touchforms/formplayer/api.py
```diff
@@ -293,7 +295,6 @@ def post_data_helper(d, auth, content_type, url, log=False):
 
 
 def formplayer_post_data_helper(d, auth, content_type, url):
-    d['nav_mode'] = 'prompt'
```
gcapalbo: looks like nav_mode won't be set to prompt for any domains using the experiment. is that intentional?

wpride: ... damnit, thanks Gio

touchforms/formplayer/sms.py
```diff
@@ -22,28 +25,28 @@ def start_session(config):
...
+    xformsresponse = answer_question(session_id, _tf_format(answer), domain, auth)
+    for resp in _next_responses(xformsresponse, session_id, domain, auth):
         yield resp
 
-def _next_responses(xformsresponse, session_id, auth=None):
+def _next_responses(xformsresponse, session_id, domain, auth=None):
     if xformsresponse.is_error:
         yield xformsresponse
     elif xformsresponse.event.type == "sub-group":
         response = next(session_id, auth)
```
gcapalbo: just realized, looks like you missed one place to pass domain

wpride:
👍

PR #306, opened 2018-06-15
Start passing hmac digest on smsforms formplayer requests

gcapalbo:
We can release this without changes to formplayer. Once formplayer validates this header in its requests then we can get rid of the Cookie header it's passing now.

Depends on https://github.com/dimagi/commcare-hq/pull/20916

@snopoke @wpride

### DIFF
--- a/touchforms/formplayer/api.py
+++ b/touchforms/formplayer/api.py
@@ -8,6 +8,7 @@ import copy
 from dimagi.utils.couch.cache.cache_core import get_redis_client
 
 from corehq.form_processor.utils.general import use_sqlite_backend
+from corehq.util.hmac_request import get_hmac_digest, convert_to_bytestring_if_unicode
 from touchforms.formplayer.exceptions import BadDataError
 from experiments import FormplayerExperiment
 from corehq import toggles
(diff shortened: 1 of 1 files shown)


touchforms/formplayer/api.py
```diff
@@ -295,21 +296,22 @@ def post_data_helper(d, auth, content_type, url, log=False):
...
 
 def formplayer_post_data_helper(d, auth, content_type, url):
-    data = json.dumps(d)
+    data = json.dumps(d).encode('utf-8')
     up = urlparse(url)
-    logging.info("Request to url: %s" % up.geturl())
     headers = {}
     headers["Content-Type"] = content_type
     headers["content-length"] = len(data)
     headers["Cookie"] = 'sessionid=%s' % settings.FORMPLAYER_INTERNAL_AUTH_KEY
```
snopoke: Could comment here that this is to be removed once formplayer supports the mac digest.

gcapalbo: updated https://github.com/dimagi/touchforms/pull/306/commits/e547a71491918437cdd2c4799203fa9232b62696

